PR #3, opened 2016-05-13
Added a setup.py and described some dependencies.

pelson:


rhattersley:
:+1: an excellent start

PR #6, opened 2016-05-13
Importable

pp-mo:
Moved source + fixed imports make a barely importable, functional module.


.gitignore
```diff
@@ -0,0 +1,29 @@
...
+.installed.cfg
+
+# Installer logs
+pip-log.txt
+
+#shared objects
+*.so
+
+#config
+cf_units/etc/site.cfg
```
pelson: Probably not necessary 😉

pp-mo: Ooh I see.
I just copied the cf_utils one.
I hadn't realised that would show up so clearly !

iris_grib/__init__.py
```diff
@@ -1 +1,1098 @@
+# (C) British Crown Copyright 2010 - 2016, Met Office
+#
+# This file is part of Iris.
```
pelson: Not yet, but we will need to decide on this (and whether we need a header at all).

.gitignore
rhattersley: I'd like to test the doc building ... so please can we fix this in a follow-up PR.

PR #9, opened 2016-05-13
Added first test, tests support.

pp-mo:
Add some basic testing support and one test, supposedly working...


iris_grib/__init__.py
```diff
@@ -36,11 +37,16 @@
...
 import scipy.interpolate
+import threading
+import warnings
 
 import iris
 import iris.fileformats.rules as iris_rules
-from iris._deprecation import warn_deprecated
-from iris.analysis._interpolate_private import Linear1dExtrapolator
+try:
+    from iris.analysis._interpolate_private import Linear1dExtrapolator
```
pelson: We *definitely* need a comment here if it is to remain in the PR...

pelson:
Tempted to merge even though `iris_grib.FUTURE.strict_grib_load` is not necessarily going to be the necessary mechanism.

rhattersley:
> Tempted to merge even though iris_grib.FUTURE.strict_grib_load is not necessarily going to be the necessary mechanism.

Let's ditch the whole FUTURE thing.

pp-mo:
Now works again, drastically reduced, hopefully now complies with "Let's ditch the whole FUTURE thing."

This currently defines it's own placeholder for 'warn_deprecated', as it doesn't exist in Iris 1.9.
That can be **removed** when we have ditched all the deprecated functionality.
Ok @rhattersley ?

rhattersley:
Close enough. Let's make progress...

Issue #8, opened 2016-05-13
Remove deprecations


rhattersley:
~~I.e. where it currently states ".. deprecated:: 1.6" or whatever, we should change to the relevant iris-grib version.~~

No deprecations!

rhattersley:
Watch out for [auto]_regularise. :unamused:

PR #10, opened 2016-05-16
Assign all initial deprecations to version 1.0.

pp-mo:
Addresses #8 
As the code parallels the existing iris.fileformats.grib, the API should be complete, consistent and stable.  So our first released version may as well be called "1.0".


rhattersley:
We don't want _any_ deprecations in the release. The "old" code should be hidden. The "new" code + @marqh's "GRIB 1 fallback" thing should be the only choice.

pp-mo:
> @rhattersley We don't want any deprecations in the release

Agreed - we are not doing this now, at all

PR #11, opened 2016-05-16
Travis tests

pp-mo:
Attempt to enable testing in Travis.
I dropped nosetests requirement for now as I don't know of a need for it, and it makes this slightly simpler.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,23 @@ install:
         rm -rf iris_source;
      fi
 
+  # Download iris-test-data
+  # --------------------------------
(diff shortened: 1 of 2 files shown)


rhattersley:
Rebase?

pp-mo:
Rebased, testing
(but probably needs a sqwish)

pp-mo:
Now working, rebased + sqwished.

NOTE: it now downloads iris-test-data in Travis.
I think this is worth doing, so we can quickly migrate existing tests from Iris to here.

rhattersley:
:+1: for getting them working, but wow, that's a lot of noise from a single test! :scream:

PR #13, opened 2016-06-02
Added specialised IrisGribTest.get_result_path

bjlittle:
@pelson as per https://github.com/SciTools/iris/pull/2037

### DIFF
--- a/iris_grib/tests/__init__.py
+++ b/iris_grib/tests/__init__.py
@@ -34,21 +34,24 @@ import os.path
 from iris.tests import IrisTest, main, skip_data, get_data_path
 
 
-#: Basepath for test results.
+#: Basepath for iris-grib test results.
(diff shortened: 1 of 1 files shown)


iris_grib/tests/__init__.py
```diff
@@ -84,3 +87,6 @@ def result_path(self, basename=None, ext=''):
                               self.__class__.__name__.replace('Test_', ''),
                               filename)
         return result
+
+
+get_result_path = IrisGribTest.get_result_path
```
pelson: Let's not add this.

pelson:
👍

iris_grib/tests/__init__.py
bjlittle: Agreed.

bjlittle:
@pelson Travis may barf for the `python=27 & iris=1.9` matrix case, but `python=27 & iris=master` should be okay ...

PR #23, opened 2016-06-20
Unit tests for iris_grib.message

bjlittle:
Porting `iris.tests.unit.fileformats.grib.message` to `iris_grib`


bjlittle:
@dkillick Fancy taking this on ...

PR #24, opened 2016-06-21
Unit tests for iris_grib._save_rules.

bjlittle:


iris_grib/tests/unit/save_rules/__init__.py
```diff
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2013 - 2015, Met Office
+# (C) British Crown Copyright 2013 - 2016, Met Office
```
dkillick: Are these dates correct? Sure, we've had the code since 2013, but it's been Iris code for most of that...

iris_grib/tests/unit/save_rules/__init__.py
```diff
@@ -14,22 +14,26 @@
...
+import iris_grib.tests as tests
+
+import numpy as np
 
 import iris
 from iris.fileformats.pp import EARTH_RADIUS as PP_DEFAULT_EARTH_RADIUS
-from iris.tests import mock
-import numpy as np
+
+from iris_grib.tests import mock
```
dkillick: Why are you getting mock from the tests and not importing directly?

iris_grib/tests/unit/save_rules/__init__.py
bjlittle: The git history has been pulled from `iris` into `iris_grib` so I've honoured the dates here and everywhere else ... just simply updated to this year.

iris_grib/tests/unit/save_rules/__init__.py
bjlittle: I've used the same pattern here as in `iris` ... nothing new really.

Happy to refactor everything if you feel strongly enough ...

iris_grib/tests/unit/save_rules/test_identification.py
```diff
@@ -14,25 +14,26 @@
...
 # importing anything else.
-import iris.tests as tests
+import iris_grib.tests as tests
 
 import gribapi
 
-import iris.fileformats.grib
-from iris.fileformats.grib._save_rules import identification
-from iris.tests import mock
+import iris
```
dkillick: I'm not a fan of how much we're having to import Iris in these tests. It makes them a little too much like integration tests!

bjlittle: It's using `iris.coords.AuxCoord` to supplement an iris stock cube. I wouldn't class that as an integration test.

But I agree that we should be careful.

dkillick:
@bjlittle merge once you have the green Travis light!

PR #26, opened 2016-06-22
Relocate iris_grib to lib/iris_grib

bjlittle:


bjlittle:
@dkillick Fancy taking this on ...

PR #27, opened 2016-06-22
Test coverage with coveralls.

bjlittle:
Added test coverage loveliness.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,16 @@
+#
+# .coveragerc to control coverage.py
+#
+
+[run]
+branch = True
+omit = 
+    setup.py
+    iris_grib/tests/*
+
+
+[report]
+exclude_lines =
+    pragma: no cover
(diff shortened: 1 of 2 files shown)


bjlittle:
@dkillick Fancy looking at this ...

.coveragerc
```diff
@@ -0,0 +1,16 @@
+#
+# .coveragerc to control coverage.py
+#
+
+[run]
+branch = True
+omit = 
+    setup.py
+    iris_grib/tests/*
```
dkillick: Really?

bjlittle: Yes, really. The point of testing coverage is to determine the percentage of code that is tested. 

By running the tests, coverage calculates the coverage of code that is exercised by running the tests. We don't want to dilute that statistic with the actual test code itself.

dkillick:
Like it! :+1: for test coverage :smile:

PR #29, opened 2016-06-23
Remove deprecated and unused reset_load_rules.

bjlittle:

PR #30, opened 2016-06-23
Remove deprecated as_pairs function.

bjlittle:


iris_grib/__init__.py
```diff
@@ -1028,19 +1028,6 @@ def save_grib2(cube, target, append=False, **kwargs):
     save_messages(messages, target, append=append)
 
 
-def as_pairs(cube):
-    """
-    .. deprecated:: 1.10
```
dkillick: Ooh, this is a little more tricky: Iris v1.10 is the current version so can we say this has observed its deprecation period?

Answers on a postcard :postbox:

dkillick:
Again, relates to #8, I'm guessing?

dkillick:
@bjlittle you've got merge conflicts now that I've merged #29...

iris_grib/__init__.py
bjlittle: My take on it is that it's deprecated in `iris` as a fall-back implementation. `iris-grib` on the otherhand doesn't need to honour that deprecation. It's a brand new world, so purge the deprecations as per #8

dkillick: Works for me!

bjlittle:
@dkillick All rebased and lookin' good ... :sunglasses:

dkillick:
Excellent! More deprecations squashed :bug: Thanks @bjlittle

PR #41, opened 2016-07-13
Fix grib wrapper edition 1 test.

bjlittle:
Hmm don't quite understand how this came about, but here is the patch to fix master :confused:


bjlittle:
@marqh Good to go :wink:

PR #42, opened 2016-07-13
Remove inherited Iris deprecations.

bjlittle:
Closes https://github.com/SciTools/iris-grib/issues/8


bjlittle:
@marqh Death to deprecation's. Long live deprecation's! :tada:

PR #45, opened 2016-07-18
Defer import of iris to avoid circularity.

pp-mo:
Patching this into iris as a direct replacement for ```iris.fileformats.grib``` will currently cause a circular import.
Because :
 * importing ```iris``` imports ```iris.fileformats```
 * importing ```iris.fileformats``` imports ```iris_grib```
 * importing ```iris_grib``` imports ```iris.fileformats.rules```  (see existing code below)

N.B. we think this probably also means we shouldn't regard iris as a _dependency_ of iris_grib.
Maybe an optional dependency ?


pp-mo:
ping @marqh

PR #46, opened 2016-07-22
pdt8 sigRefTime

marqh:
to fix potential bug when significance of reference time indicates a time, thus a time bounds, within PDT8


marqh:
ping @pp-mo

PR #51, opened 2016-08-05
Add save plumbing for pdt10

### LINKED ISSUE #50: Save processing for PDT10 not being called
There's a problem with the change added by #44: the function for [processing pdt10 for saving](https://github.com/SciTools/iris-grib/blob/master/iris_grib/_save_rules.py#L875-L891) has not been plumbed in to the [product definition section function](https://github.com/SciTools/iris-grib/blob/master/iris_grib/_save_rules.py#L1003), so it will never be executed.

This does hint at the fact that we could do with some integration tests for loading and saving of grib files using iris_grib, as there's no way a unit test framework could ever be expected to pick up on this...

dkillick:
Adds a call for the save rules function for PDT10 to `_save_rules.product_definition_section`. This ensures that the PDT10 save rules function is called when a GRIB message described by PDT10 is passed to the saver.

Corrects an omission from #44 and fixes #50.

Note: does not (currently) contain the tests I suggested in #50 that would be beneficial to ensuring such omissions are harder to make in future.

### DIFF
--- a/iris_grib/_save_rules.py
+++ b/iris_grib/_save_rules.py
@@ -797,33 +797,38 @@ def _cube_is_time_statistic(cube):
     """
     Test whether we can identify this cube as a statistic over time.
 
-    At present, accept anything whose latest cell method operates over a single
-    coordinate that "looks like" a time factor (i.e. some specific names).
-    In particular, we recognise the coordinate names defined in
-    :py:mod:`iris.coord_categorisation`.
+    We need to know whether our cube represents a time statistic. This is
+    almost always captured in the cell methods. The exception is when a
+    percentage statistic has been calculated (i.e. for PDT10). This is
+    captured in a `percentage_over_time` scalar coord, which must be handled
+    here too.
 
     """
-    # The *only* relevant information is in cell_methods, as coordinates or
-    # dimensions of aggregation may no longer exist.  So it's not possible to
-    # be definitive, but we handle *some* useful cases.
-    # In other cases just say "no", which is safe even when not ideal.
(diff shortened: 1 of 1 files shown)


dkillick:
Turns out that the first commit was not all the story, as cubes matching to PDT10 were not being recognised as time statistic cubes. This meant that the code added in the first commit was still not being run.

The second commit above improves the time statistic checking performed by `_cube_is_time_statistic` so that the particulars of cubes matching to PDT10 are interpreted correctly.

iris_grib/_save_rules.py
```diff
@@ -1006,6 +1011,7 @@ def product_definition_section(cube, grib):
     grib message based on the properties of the cube.
 
     """
+    import pdb; pdb.set_trace()
```
marqh: i think you left this in by mistake

dkillick: Bad one! I'll remove that...

iris_grib/_save_rules.py
```diff
@@ -797,33 +797,38 @@ def _cube_is_time_statistic(cube):
...
+            coord_name = latest_coordnames[0]
+            result = coord_name in recognised_time_names
+    else:
+        result = False
 
-    # Accept it if the name is recognised.
-    # Currently does *not* recognise related names like 'month_number' or
-    # 'years', as that seems potentially unsafe.
-    return coord_name in recognised_time_names
+    return result
```
marqh: the refactoring of these tests makes a lot of sense to me, I think it represents a good approach

PR #52, opened 2016-08-08
User documentation

### LINKED ISSUE #49: Public docs
The github docs link is not yet viable.
It would be nice to be able to link to this from Iris docs.

dkillick:
Adds some much-needed documentation, so that prospective users of `iris-grib` have some idea of how to use `iris-grib` to load from and save to GRIB2 files.

As an added bonus, the reference documentation hook now also works!

Oh, this also fixes #49.


marqh:
a great improvement, thanks @dkillick

PR #53, opened 2016-08-11
Api privacy

pp-mo:
Tidy what things are public in the API.
This is a sensible preliminary to providing complete documentation via Sphinx.

**NOTE** checked that iris tests still run :ok_hand:


iris_grib/__init__.py
```diff
@@ -50,7 +50,7 @@
 __version__ = '0.9.0'
 
 __all__ = ['load_cubes', 'save_grib2', 'load_pairs_from_fields',
-           'save_pairs_from_cube', 'save_messages', 'GribWrapper']
```
dkillick: Do we know why `GribWrapper` was in the public API but `GribMessage` wasn't?

Further, if we've removed it from the public API do we want to mark it as "private" with a leading underscore (_) on the class name?

iris_grib/tests/unit/grib_phenom_translation/test_grib_phenom_translation.py
```diff
@@ -31,11 +31,12 @@
 import cf_units
 
 import iris_grib.grib_phenom_translation as gptx
+from iris_grib.grib_phenom_translation import _LookupTable
```
dkillick: What benefit did splitting this particular import out provide?

iris_grib/__init__.py
pp-mo: > Do we know why GribWrapper was in the public API but GribDataProxy wasn't?

Not really, but the original had got a bit untidy.
Before the days of the 'new strict loader', GribWrapper was what callbacks had to deal with, but now it is replaced by GribMessage in that context.

The only place a GribWrapper still rears its ugly head is when loading GRIB1 messages, which we continue to support because some standard datastreams still use them (actually, mixed GRIB1 and GRIB2 -- yuck!).
The user may possibly still care because a GribWrapper will then be passed into a load callback.
See : #56


> Further, if we've removed it from the public API do we want to mark it as "private" with a leading underscore (_) on the class name?

I don't think this is necessary, as the ```__all__ ``` should be an accepted mechanism of defining what is public.
According to [this section of PEP8](https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces).
( which is also now directly referenced by https://github.com/SciTools/iris/blob/master/docs/iris/src/whitepapers/change_management.rst )

dkillick:
@pp-mo just a couple of queries from me otherwise I'm happy with this change :+1:

iris_grib/__init__.py
dkillick: >  the `__all__` should be an accepted mechanism of defining what is public

I thought you might say that, and I'm fine with that as the answer!

iris_grib/tests/unit/grib_phenom_translation/test_grib_phenom_translation.py
pp-mo: Well, not so sure now...
I guess it makes it clear that we're diddling with private elements ?

I will remove if you prefer.

dkillick: Or you could keep the `from ... import` style and add the only other import (`grib1_phenom_to_cf_info`) to the import line.

dkillick:
A nice tidy - thanks @pp-mo! Feel free to give #55 a rebase now.

PR #54, opened 2016-08-11
Tweak conda dependencies for readthedocs build.

pp-mo:
Produces readthedocs output.
( but actual docs content still needs improving )


dkillick:
Yes building docs! Thanks @pp-mo :smile:

PR #55, opened 2016-08-11
Full sphinx docs

pp-mo:
Get sphinx docs for the whole api.

The results should good enough for a first release : **shout if they're not...**

**DO NOT MERGE**
Incorporates #53 and #54.  Possibly needs rebasing when those are in.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -2,4 +2,4 @@ channels:
     - scitools
 dependencies:
     - iris=1.*
-    - grib_api=1.14.7
+    - ecmwf_grib=1.14
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -106,6 +106,12 @@ pygments_style = 'sphinx'
 todo_include_todos = False
 
 
+# -- Autodoc ------------------------------------------------------------------
+
+autodoc_member_order = 'groupwise'
+autodoc_default_flags = ['show-inheritance']
+
+
 # -- Options for HTML output ----------------------------------------------
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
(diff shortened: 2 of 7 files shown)


pp-mo:
**NOTE**
At present, you can view the docs this produces here :
http://pp-iris-grib.readthedocs.io/en/latest/index.html

environment.yml
```diff
@@ -1,4 +1,6 @@
 channels:
-    - conda-forge
```
dkillick: These changes would appear to be a duplicate of those in #54 - I guess this is why you're concerned about the merge issues?

dkillick:
Hey @pp-mo is this a combination of #53 and #54 with some extra changes also thrown in??

pp-mo:
> @dkillick is this a combination of #53 and #54 with some extra changes also thrown in??

Yes it is.
Hence "DO NOT MERGE" warning : It might need rebasing.

pp-mo:
Rebased - seems ok

environment.yml
```diff
@@ -2,4 +2,4 @@ channels:
     - scitools
 dependencies:
     - iris=1.*
-    - grib_api=1.14.7
+    - ecmwf_grib=1.14
```
dkillick: Does this need to be constrained as far as v1.14.7 still?

pp-mo: I would say no, minor version is being too picky.
If it works now it will work I think, as 1.16 is already out there, but we can't work with it for some tiresome backwards-incompatibility reasons (yet to be determined).
While I'm thinking of it... #57 #58

pp-mo:
Ping @dkillick @marqh 
This can now be merged, since #53 and #54 have gone in.
Is this ready to go too ?
It would be nice to have docs on the repo to accompany the iris 1.10 release.

P.S. we will also need to enable docs building on this repo #60.

dkillick:
I like it, thanks @pp-mo!

PR #59, opened 2016-08-12
Add section on connectivity to Iris

dkillick:
As Iris will automatically use iris-grib if it can, it's worth a little section on this within the docs.


pp-mo:
Good stuff !
Thanks @dkillick

PR #66, opened 2016-12-01
Correctly populate setup.py package_data.

bjlittle:
The `setup.py` is not correctly specifying the `package_data` for `iris_grib` i.e. the `results` of the `iris_grib.test` is missing.


bjlittle:
Thanks @dkillick !

PR #68, opened 2016-12-01
Fix interpretation of significanceOfReferenceTime=0.

### LINKED ISSUE #67: Interpretation of significanceOfReferenceTime=0 incorrect
I've just been tracking down an issue where I believe ``significanceOfReferenceTime = 0;`` is being incorrectly interpreted. According to the [`TIME RANGE INDICATOR` docs](http://apps.ecmwf.int/codes/grib/format/grib1/ctable/5/) I would interpret a significanceOfReferenceTime as a `forecast_reference_time`, not a validity `time`.

This change was introduced in iris at https://github.com/SciTools/iris/pull/1903/files#diff-e831265b91970d39925897dda02564a6R275 (ping @marqh).

pelson:
Closes #67.

### DIFF
--- a/iris_grib/_load_convert.py
+++ b/iris_grib/_load_convert.py
@@ -277,7 +277,7 @@ def reference_time_coord(section):
 
     """
     # Look-up standard name by significanceOfReferenceTime.
-    _lookup = {0: 'time',
+    _lookup = {0: 'forecast_reference_time',
                1: 'forecast_reference_time',
                2: 'time',
                3: 'time'}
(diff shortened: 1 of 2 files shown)


marqh:
hello @pelson 

i am trying to dig through my change record to find the reasoning for this change.  It was a deliberate act, so i presume there was some cause.  I have not yet found the file I was using in my investigation, which I assume had a indicator value of 0

The GRIB specification defines
```
Code table 1.2 – Significance of reference time
Code figure   Meaning
0 Analysis
1 Start of forecast
2 Verifying time of forecast
3 Observation time

```

It seemed clear that `Start of Forecast` is a good map to `forecast_reference_time`.  It is not quite so clear what the difference between `Start of Forecast` and `Analysis` is.

CF standard names states
```
forecast_reference_time
The forecast reference time in NWP is the "data time", the time of the analysis from which the forecast was made.
```
which would support the use of `forecast_reference_time` for `Analysis`, perhaps over `Start of forecast`

In summary, I think this makes sense.

pelson:
Thanks @marqh.

PR #72, opened 2016-12-06
Remove use of threads.local as a means of storing options.

### LINKED ISSUE #71: Dask worker fails to open badly translated grib file with Iris
First, the traceback:

```
  File "/data/pelson/conda/envs/data-transform/lib/python2.7/site-packages/iris_grib/_load_convert.py", line 1634, in generating_process
    if options.warn_on_unsupported:
AttributeError: 'thread._local' object has no attribute 'warn_on_unsupported'
```

----

There is some weird interaction between the pickled function and the way the function is called with dask. Essentially, I think iris-grib is relying on the _load_convert having been run, but that isn't necessarily the case when pickling/unpickling the function (or at least, the local it is referencing isn't the one that the thread has run the module on). This is a somewhat woolly description :stuck_out_tongue_winking_eye:.

Rather than depending on the module having been run, we either need to ensure that we validate the [threading.local](https://docs.python.org/2/library/threading.html#threading.local) has been correctly configured, or just ditch the complexity. I don't currently see the usecase for having warn_on_unsupported be defined on a per-thread basis. The implementation of this was originally done in https://github.com/SciTools/iris/commit/0248bf3abaf8d2ba77cc0c3b82cea23e87753ef5, and for all of those tests, I believe a module level constant would suffice.

pelson:
Closes #71.
This is the least intrusive change I can come up with that addresses the problem of threads.local usage. ``threads.local`` in itself is not a bad thing - it just needs to be initialised appropriately. That would be a more significant change here (and would require some care to avoid treading on settings that a user wishes to override, not to mention the performance cost).

### DIFF
--- a/iris_grib/_load_convert.py
+++ b/iris_grib/_load_convert.py
@@ -23,10 +23,10 @@ cube metadata.
 from __future__ import (absolute_import, division, print_function)
 from six.moves import (filter, input, map, range, zip)  # noqa
 
+from argparse import Namespace
 from collections import namedtuple, Iterable, OrderedDict
 from datetime import datetime, timedelta
 import math
-import threading
 import warnings
 
 import cartopy.crs as ccrs
@@ -49,9 +49,9 @@ from .message import GribMessage
 # Restrict the names imported from this namespace.
 __all__ = ['convert']
 
-options = threading.local()
-options.warn_on_unsupported = False
-options.support_hindcast_values = True
+
+options = Namespace(warn_on_unsupported=False,
+                    support_hindcast_values=True)
 
 ScanningMode = namedtuple('ScanningMode', ['i_negative',
                                            'j_positive',


bjlittle:
@pelson Is `argparse.Namespace` thread-safe?

pelson:
> Is argparse.Namespace thread-safe?

No. Does it need to be?

pelson:
P.S. Go to bed :sleeping_bed:  :smile:

pelson:
I'm just running to lunch :plate_with_cutlery:, will be back in an hour :smile:

bjlittle:
Can't sleep :sleeping: 

... consider the case of threaded nose tests that change these attribute values. One thread changing state shouldn't influence any other ...

pelson:
> One thread changing state shouldn't influence any other

Agreed. But nose doesn't do tests in threads... 
Because of the GIL there really is no benefit in most cases for running python unittests threadded (typically we avoid doing heavily I/O, or CPU intensive operations on large numpy arrays in tests).

My belief is that the benefits of making iris-grib thread-safe are not worth the effort/maintenance burden. Particularly as I believe we already have an iris-grib that isn't *actually* functional in a multi-process environment when combined with pickle/cloudpickle (#71).

pelson:
I decided to be a bit less woolly about my description in #71. I've put together a demonstration of how the code is already absolutely *not* thread-safe (because it doesn't work in anything but the main thread):

```
sub_mod = """
import threading

options = threading.local()
options.foo = 'bar'

def do_your_thang():
    print('Howdy from {}. Your options {} foo'
          ''.format(threading.current_thread(),
                    'include' if hasattr(options, 'foo') else "don't include"))

"""

with open('my_submod.py', 'w') as fh:
    fh.write(sub_mod)

import my_submod

import concurrent.futures

# See ma, it works!
my_submod.do_your_thang()

with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
    r = executor.submit(my_submod.do_your_thang)
    # Wait, it doesn't work after all! :(
    r.result()
```

Result:

```
python threaded_bad.py 
Howdy from <_MainThread(MainThread, started 140175739709248)>. Your options include foo
Howdy from <Thread(Thread-1, started daemon 140175583192832)>. Your options don't include foo
```

In summary, we are maintaining code complexity for something that doesn't even work outside of the main thread. It would be easy enough to make it work (by setting up options before each call to the function), but the argument that there is virtually no benefit to making this code thread-safe comes into play.

pelson:
@bjlittle - further thoughts on this. It is a problem that anybody wanting to use dask (distributed) with iris-grib will face.

marqh:
this looks pretty non-controversial to me

I'm happy to take this as it

thank you @pelson

PR #73, opened 2017-03-08
Add support for Lambert Conformal saving

djkirkham:
No tests yet, will add them shortly.

False eastings/northings are not supported for Lambert Conformal in GRIB, so they are ignored on save. Doing this doesn't lose any information about the locations of the points though, since GRIB stores the longitude and latitude of the first point in the grid.

### DIFF
--- a/iris_grib/_save_rules.py
+++ b/iris_grib/_save_rules.py
@@ -527,15 +527,15 @@ def grid_definition_template_30(cube, grib):
     first_x = first_x % 360
     central_lon = cs.central_lon % 360
 
-    gribapi.grib_set_long(grib, "latitudeOfFirstGridPoint",
-                          int(first_y * 1e6))
-    gribapi.grib_set_long(grib, "longitudeOfFirstGridPoint",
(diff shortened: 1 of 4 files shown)


djkirkham:
Thinking about it, it's probably worth at least having a warning that the false eastings/northings won't be saved.

lbdreyer:
It might be worth adding a loader for Lambert Conformal at the same time. Then you can check what happens when you load in and save out a Lambert Conformal, and compare that to the original.

djkirkham:
@lbdreyer There already is a loader. It saves the correct values in the file, but the issue is with saving and then loading, where the false eastings/northings are lost. I'm not sure if this should be considered a problem.

djkirkham:
I still haven't added testing code, so don't merge yet

marqh:
@djkirkham any further tests planned, or is this good to go from your point of view?

djkirkham:
@marq It's ready to be merged from my point of view

marqh:
I am also happy with this

thank you @djkirkham

PR #75, opened 2017-03-16
Prevent usage of floats as indices to numpy arrays due to deprecation

djkirkham:
Newer versions of numpy don't allow floats to be used as indices to arrays, even if they are integer-valued. This PR makes the tests pass (plus iris integration tests) with newer version of numpy, but I have not assessed other places in the code which may cause errors but are not caught by tests.


pelson:
I don't think this is the right thing to do here. XRef https://github.com/SciTools/iris/pull/2468.

pelson:
Though I confess - I have no idea what the author was aiming for with the zeros use to determine the system float type (which can be got with ``np.float_``).

PR #76, opened 2017-03-24
Support for pdt 4.15.

pp-mo:
Only supports spatial processing code 0 = "whole cell, no interpolation".


marqh:
this looks fine, thank you @pp-mo

PR #77, opened 2017-03-27
Product Definition Template 32 (synthetic satellite data)

djkirkham:
Add support for loading Product Definition Template 32.

### DIFF
--- a/iris_grib/__init__.py
+++ b/iris_grib/__init__.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2016, Met Office
+# (C) British Crown Copyright 2010 - 2017, Met Office
 #
 # This file is part of iris-grib.
 #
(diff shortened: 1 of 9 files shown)


pp-mo:
It felt to me like we should have a separate test for the 'satellite_common' routine.

Rather than just asking, I've proposed some changes, here : https://github.com/djkirkham/iris-grib/pull/1
Note: I think all tests pass there, but you haven't got Travis enabled on your repo.

In there, I have also added some other tidies:
  * rationalised how it issues the generating-process warnings, by extending usage of "generating_process"
  * avoided checking for warnings in all the derived templates
  * factored out the repeated code relating to multiple wave-number values and scale factors

Much more that could be done along these lines, but I hope this is a small + contained improvement.

pp-mo:
P.S. otherwise all is good -- the implementation looks ok to me.

One really tiny point : I'm not sure that the naming of "satellite_common" is in line with the other routines in `_load_convert`.  They seem to be just named for the aspect they deal with, so maybe something like 'satellite_wavebands' would be more appropriate ?

marqh:
looks good, thank you @djkirkham

PR #78, opened 2017-03-27
Allow dynamically supporting individual templates

djkirkham:
In the current implementation, adding support for a particular Product or Grid Definition Template requires a release of the library. It would be nice to be able to add support for a template at runtime. This is not possible with an `iris.load()` callback because the load fails before the callback is even reached.

This PR demonstrates how the code could be changed to allow adding a function to handle a particular template at runtime. Currently it still requires accessing a private module.

```
import iris
from iris_grib._load_convert import ProductDefinitionTemplateHandlers

def my_handler(section, metadata, rt_coord):
    ...

ProductDefinitionTemplateHandlers[32] = my_handler

# Load a file which uses Product Definition Template 32
iris.load('myfile.grib2')
```

### DIFF
--- a/iris_grib/_load_convert.py
+++ b/iris_grib/_load_convert.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2014 - 2016, Met Office
+# (C) British Crown Copyright 2014 - 2017, Met Office
 #
 # This file is part of iris-grib.
 #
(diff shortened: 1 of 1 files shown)


pp-mo:
This has great promise for providing user patches, but I think it needs a clear statement as to how + when you can reasonably use it.

New template code that will be invoked by this will generally be full of calls to other conversion routines in `_load_convert`.   Those are all private, and I believe will _never_ be public (supported) API : we need to freely rewrite those and their inter-relations as we add different templates (e.g. see recent PRs #76 and #77).

So, this lets you add a workaround patch, but such solutions are always pretty hacky and unstable.
It really only makes sense for 'pre-releasing' code that will be added in a later release, and can't be expected to work for anything other than the release it was designed for.

A patch could simply fail at some subsequent release, but it could also potentially introduce a subtle bug if the underlying code is modified without reference to user 'patch' code that calls it.
We _could_ avoid that by consistently renaming anything routine we modify, or making patch code explicitly version-specific ?

pp-mo:
*NOTE:* this should now probably be recast as an Iris PR.
( Now we are re-integrating iris grib support within the main repo : https://github.com/SciTools/iris/pull/2486 )

djkirkham:
@pp-mo Thanks for the comments.

> New template code that will be invoked by this will generally be full of calls to other conversion routines in _load_convert

Well, you could simply copy and paste the code needed into the patch itself. I know it's not pretty...

> it could also potentially introduce a subtle bug if the underlying code is modified without reference to user 'patch' code that calls it.

If the patch code only used public methods, and had any private code it needed copied, this needn't be a problem. The only issue would be that once the correct template was added, the patch code would fail with a key error, but you could get around this with a try-except in the patch code.

PR #82, opened 2017-06-25
Port the changes from iris.fileformats.grib back into iris_grib.

pelson:
I've done this with very little understanding of the original rationale for integrating iris_grib back into iris, so I may have missed a key detail.

### DIFF
--- a/iris_grib/message.py
+++ b/iris_grib/message.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2014 - 2016, Met Office
+# (C) British Crown Copyright 2014 - 2017, Met Office
 #
 # This file is part of iris-grib.
 #
@@ -26,10 +26,11 @@ import six
 from collections import namedtuple
 import re
 
-import biggus
 import gribapi
 import numpy as np
+import numpy.ma as ma
 
+from iris._lazy_data import array_masked_to_nans, as_lazy_data
 from iris.exceptions import TranslationError
 
 
@@ -88,7 +89,7 @@ class GribMessage(object):
         """
         # A RawGribMessage giving gribapi access to the original grib message.
         self._raw_message = raw_message
-        # A _MessageLocation which biggus uses to read the message data array,
+        # A _MessageLocation which dask uses to read the message data array,
         # by which time this message may be dead and the original grib file
         # closed.
         self._recreate_raw = recreate_raw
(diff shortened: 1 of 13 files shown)


pelson:
Note: I accidentally pushed this branch to scitools/iris-grib, rather than my own fork. Apologies for that.

pelson:
For the record, the pertinent commit in Iris that brought iris_grib back in was: https://github.com/SciTools/iris/commit/e45acf7597c0aeeaf0cd0d8e87503de175a3fa99#diff-3967991fd23406c94b677376dc007fd8

iris_grib/message.py
```diff
@@ -180,12 +194,13 @@ def __call__(self):
 class _DataProxy(object):
     """A reference to the data payload of a single GRIB message."""
 
-    __slots__ = ('shape', 'dtype', 'fill_value', 'recreate_raw')
+    __slots__ = ('shape', 'dtype', 'recreate_raw')
 
     def __init__(self, shape, dtype, fill_value, recreate_raw):
+        # TODO: I (@pelson) have no idea why fill_value remains an argument as
```
pelson: This looks like an oversight in iris.

lbdreyer:
> I've done this with very little understanding of the original rationale for integrating iris_grib back into iris

You have been shown this before but in case you forgot, there was an attempt to open the discussion about the future of iris_grib [on the goole groups](https://groups.google.com/forum/#!topic/scitools-iris-dev/lFw2_CnaEHU).

I don't understand the purpose of this PR. As far as I'm concerned iris_grib is dead

pelson:
Thanks @lbdreyer. The link was helpful - I've written something there, and suggested we get together to discuss iris' direction in a slightly higher-bandwidth form. (doodle poll for date/time: https://doodle.com/poll/yaxm4pqfkus96sm2)

Suffice to say, I personally do not feel that the subject has been considered sufficiently to call iris-grib a dodo just yet...

pelson:
The thread was: https://groups.google.com/forum/#!topic/scitools-iris-dev/lFw2_CnaEHU
The direction is for SciTools/iris-grib to be the canonical GRIB loading/saving capability for Iris.

In light of this, could this please get a review? @SciTools/iris-grib-devs

pp-mo:
I'm happy that this is equivalent to code in Iris at "version 2.0a0", i.e. https://github.com/SciTools/iris/commit/7ec22fa3f14b1aea454c5a51ac350de190603d9c.
There is a small code difference between the `load_convert.ellipsoid` routines, between use of `if` and `elif` : 
 ( compare : https://github.com/SciTools/iris-grib/blob/6c9fb54b01eeac9ebeeec4a5337223159b31efae and /iris_grib/_load_convert.py#L387 )

Note that this version *only* works with current master, and will not work for long,
because the 'dask nanmask solution' code is ephemeral + will not be available in a tag or release.
So, we need to fix again soon wrt the imminent re-merge of the feature branch https://github.com/SciTools/iris/tree/dask_mask_array

PR #86, opened 2017-10-17
Fix for backwards-incompatible change in iris.fileformats.rules.Loader

pp-mo:
Simple fix, just to prove recent merge is o.k.

It may not be worth much to merge this, as https://github.com/SciTools/iris/tree/dask_mask_array is coming soon ...


pp-mo:
> Where does this PR sit. Does it need merging?

In principle this change will be needed.
But since https://github.com/SciTools/iris/pull/2824 a whole lot of iris-grib will need changing anyway.

pp-mo:
This commit now subsumed in #87

PR #87, opened 2017-10-24
Dask masked fix

pp-mo:
Bring iris_grib up-to-date with recent changes to Iris, 
to use masked-arrays instead of nan masking in Dask (Dask v 0.13).

### DIFF
--- a/iris_grib/__init__.py
+++ b/iris_grib/__init__.py
@@ -36,7 +36,7 @@ import numpy as np
 import numpy.ma as ma
 
 import iris
-from iris._lazy_data import as_lazy_data, convert_nans_array
+from iris._lazy_data import as_lazy_data
 import iris.coord_systems as coord_systems
 from iris.exceptions import TranslationError, NotYetImplementedError
 
@@ -145,7 +145,6 @@ class GribWrapper(object):
     def __init__(self, grib_message, grib_fh=None):
         """Store the grib message and compute our extra keys."""
         self.grib_message = grib_message
-        self.realised_dtype = np.array([0.]).dtype
 
         if self.edition != 1:
             emsg = 'GRIB edition {} is not supported by {!r}.'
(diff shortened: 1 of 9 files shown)


iris_grib/__init__.py
```diff
@@ -145,7 +145,6 @@ class GribWrapper(object):
     def __init__(self, grib_message, grib_fh=None):
         """Store the grib message and compute our extra keys."""
         self.grib_message = grib_message
-        self.realised_dtype = np.array([0.]).dtype
```
pelson: API change

iris_grib/__init__.py
```diff
@@ -698,6 +694,10 @@ def _message_values(grib_message, shape):
     data = gribapi.grib_get_double_array(grib_message, 'values')
     data = data.reshape(shape)
 
+    # Handle missing values in a sensible way.
```
pelson: Is this a hangover from not having masks?

iris_grib/_save_rules.py
```diff
@@ -1157,17 +1157,20 @@ def product_definition_section(cube, grib):
...
+            # Use the data's fill value.
+            fill_value = float(cube.data.fill_value)
+        else:
             # We can't use the cube's fill value if it's NaN,
             # the GRIB API doesn't like it.
             # Calculate an MDI outside the data range.
             min, max = cube.data.min(), cube.data.max()
             fill_value = min - (max - min) * 0.1
+        # Prepare the unmasked data array, using fill_value as the MDI.
+        data = cube.data.filled(fill_value)
```
pelson: Again, is this a hangover?

iris_grib/__init__.py
pp-mo: Well, not if we never released the previous version, no ? ;-)

iris_grib/__init__.py
pp-mo: Not totally sure.
This is simply reflecting what is currently in Iris master.

iris_grib/_save_rules.py
pp-mo: As above...

iris_grib/__init__.py
pelson: I'd love to find out more...

PR #92, opened 2017-10-27
Changes ported from Iris#2607

pp-mo:
This won't work until an Iris 2 can be found on conda-forge


pelson:
Huge thanks to @marqh for making this happen in https://github.com/SciTools/iris/pull/2607.

Thanks for converting this to iris-grib @pp-mo!

PR #100, opened 2018-05-29
Test against latest conda-forge iris.

pp-mo:
**WIP**
Test have been failing since it tests against iris-master, but was also relying on "iris/minimal-conda-requirements.txt", which is now defunct.
We "ought" to be testing against the latest Iris release, not master, in any case.
I think this was not doing so as it predated our first usable Iris 2 release.

Hopefully we now have a usable Iris v2 on conda-forge, so we should now test against that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ install:
 
   # Install iris-grib's dependencies
   # --------------------------------
-  - conda install --quiet --yes python=${python} iris ecmwf_grib;
+  - conda install --quiet --yes python=${python} iris ecmwf_grib numpy=1.13 cartopy=0.15 mock filelock pep8;
   - conda list
 
   # Download iris-test-data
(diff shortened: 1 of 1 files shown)


dkillick:
@pp-mo the code changes look ideal (though, as discussed offline, we'll probably need to add more packages to the install list) and we've moved the testing error onto a failure at least, so things are looking up! 🙄

pp-mo:
> looking up 🙄

Well it now passes !

I think this is "close", but there are some irritating dependency problems which here I'm working around with some pins ...
 * cartopy 0.16 broke Iris `coord_systems.Stereographic.as_cartopy_crs` : fixed in latest Iris but not released yet :  see https://github.com/SciTools/iris/pull/2971
 * numpy 1.13 --> 1.14 broke Iris assertCML : see https://github.com/SciTools/iris/pull/3019/files#r188917443

All this should get fixed in Iris 2.1, so I think we can wait for that rather than trying to reinvent the wheel

dkillick:
@pp-mo nice work - this looks fine (and the tests are passing! 💯) Need to remember to treat this as an intermediary and unpin when Iris 2.1 is released, but there's no point in reinventing the wheel, as you said.

PR #102, opened 2018-06-01
Save data on hybrid heights.

pp-mo:
( this should be okay now. )

It's not so simple, because of the slightly insane GRIB method of recording metadata for all hybrid levels in *every* message.  Thus, we need to pass the 'full original' cube into each translation call, not just the 2d slice we are currently dealing with, which mangles `_save_rules.py` quite a lot.

Also, for the same reason, the length of the 'pv' vector within each recorded message, and location of relevant coefficient values within it, depend on the actual model-level number of that 2d field.


iris_grib/tests/integration/save_rules/test_save_hybrid_height.py
```diff
@@ -0,0 +1,91 @@
...
+from six.moves import (filter, input, map, range, zip)  # noqa
+
+# import iris_grib.tests first so that some things can be initialised before
+# importing anything else
+import iris_grib.tests as tests
+
+
+import iris
+from iris.cube import Cube
+from iris import sample_data_path
```
corinnebosley: Apparently this isn't the right way to import the sample data.  We will have to figure this out.

pp-mo: This is not importing the sample data, only the iris 'sample_data_path' method.

The sample data would need to be installed by Travis.yml
It's pretty heavyweight + I don't think we want to start doing that for iris-grib, so I'll look for some other test data ...

iris_grib/_save_rules.py
```diff
@@ -731,13 +733,33 @@ def set_forecast_time(cube, grib):
...
+    # checkout factories for hybrid vertical coordinates.
+    hybrid_height_factories = [
+        factory for factory in cube.aux_factories
+        if isinstance(factory, HybridHeightFactory)]
+    if not hybrid_height_factories:
+        hybrid_height_factory = None
+    else:
+        # N.B. get the matching 'complete' factory from the full 3d cube, so we
+        # have all the level information.
+        factory, = hybrid_height_factories
```
corinnebosley: Your implementation above seems to suggest that we could have a list of HybridHeightFactories (although I would consider it unlikely...).  If that happens, then the unpacked variable `factory` would become a list and then not have a name.

Would this be resolved by saying:
`factory, = hybrid_height_factories[0]`?

pp-mo: > seems to suggest that we could have a list of HybridHeightFactories

That wasn't intended.
In fact, it's just a sneaky way of fetching the first of a list while simultaneously forcing an error if it doesn't have exactly one element.
It's handy, but every time I write one, I wonder if it's just a bit too clever ...

corinnebosley: Sorted.  Not a problem.

iris_grib/tests/integration/save_rules/test_save_hybrid_height.py
```diff
@@ -0,0 +1,88 @@
...
+            msg_lev2 = msgs[1]
+            #  first surface type = 118  (i.e. hybrid height).
+            self.assertEqual(
+                msg_lev2.sections[4]['typeOfFirstFixedSurface'],
+                118)
+            #  first surface scaling = 0.
+            self.assertEqual(
+                msg_lev2.sections[4]['scaleFactorOfFirstFixedSurface'],
+                0)
+            #  first surface value = 2  (i.e. model level).
```
corinnebosley: Can you please expand on `(model level)` so that I know why it's model level number 2? i.e. `(model level number coordinate of corresponding cube slice)` or something.  I had to look at the ticket (AVD-1021) to work out what it meant.

pp-mo:
Status update -- still **DO NOT MERGE** for now : 
  * still needs a response to your [call for more clarity above](https://github.com/SciTools/iris-grib/pull/102#discussion_r193016842) above
  * will be better for a  rebase after #106 fixes
  * given #107 can also now add a whatsnew

pp-mo:
Think this is now good to go ?
Ping @corinnebosley @dkillick

iris_grib/_save_rules.py
```diff
@@ -731,13 +733,34 @@ def set_forecast_time(cube, grib):
...
+    if not hybrid_height_factories:
+        hybrid_height_factory = None
+    else:
+        # If any, there should be just one.
+        factory, = hybrid_height_factories
+        # Fetch the matching 'complete' factory from the *full* 3d cube, so we
+        # have all the level information.
+        hybrid_height_factory = full3d_cube.aux_factory(factory.name())
+
+    # hybrid pressure
```
corinnebosley: This comment says 'hybrid pressure', but it's looking for a hybrid height factory with a vertical coordinate code 118, which is specifically hybrid height, not hybrid pressure.  Is it just the comment that's misplaced?

pp-mo: >  Is it just the comment that's misplaced?

Yes I think so.
Good spot :trophy:

corinnebosley:
@pp-mo Boom!

Issue #95, opened 2018-02-13
Mercator grid not supported?


lkugler:
Hi there,
I got a `TranslationError` reporting mercator is an unhandled grid type.
Does this mean that mercator grid in general is not supported by **iris-grib**?
Thanks!

```
> File "iris-plot.py", line 21, in plotit
>     cubelist = iris.load(gribpath)
>   File "/home/miniconda2/lib/python2.7/site-packages/iris/__init__.py", line 341, in load
>     return _load_collection(uris, constraints, callback).merged().cubes()
>   File "/home/miniconda2/lib/python2.7/site-packages/iris/__init__.py", line 311, in _load_collection
>     result = iris.cube._CubeFilterCollection.from_cubes(cubes, constraints)
>   File "/home/miniconda2/lib/python2.7/site-packages/iris/cube.py", line 144, in from_cubes
>     for cube in cubes:
>   File "/home/miniconda2/lib/python2.7/site-packages/iris/__init__.py", line 298, in _generate_cubes
>     for cube in iris.io.load_files(part_names, callback, constraints):
>   File "/home/miniconda2/lib/python2.7/site-packages/iris/io/__init__.py", line 201, in load_files
>     for cube in handling_format_spec.handler(fnames, callback):
>   File "/home/miniconda2/lib/python2.7/site-packages/iris/fileformats/rules.py", line 1047, in load_cubes
>     user_callback_wrapper=loadcubes_user_callback_wrapper):
>   File "/home/miniconda2/lib/python2.7/site-packages/iris/fileformats/rules.py", line 960, in _load_pairs_from_fields_and_filenames
>     for field, filename in fields_and_filenames:
>   File "/home/miniconda2/lib/python2.7/site-packages/iris/fileformats/rules.py", line 1024, in _generate_all_fields_and_filenames
>     filename, **loader.field_generator_kwargs):
>   File "/home/miniconda2/lib/python2.7/site-packages/iris_grib/__init__.py", line 700, in _load_generate
>     message = GribWrapper(message_id, grib_fh=grib_fh)
>   File "/home/miniconda2/lib/python2.7/site-packages/iris_grib/__init__.py", line 170, in __init__
>     self._compute_extra_keys()
>   File "/home/miniconda2/lib/python2.7/site-packages/iris_grib/__init__.py", line 460, in _compute_extra_keys
>     raise TranslationError("unhandled grid type: {}".format(gridType))
> iris.exceptions.TranslationError: unhandled grid type: mercator
[...]

PR #104, opened 2018-06-04
Support for loading grid_definition_template_10 (Mercator projection)

kaedonkers:
There was not support in iris-grib for loading data from grib2 files in the Mercator projection. This pull request addresses that by adding `grid_definition_template_10` to `_load_convert.py`.

Related to #95.

### DIFF
--- a/iris_grib/_load_convert.py
+++ b/iris_grib/_load_convert.py
@@ -719,6 +719,42 @@ def grid_definition_template_5(section, metadata):
                                      'grid_latitude', 'grid_longitude', cs)
 
 
+def grid_definition_template_10(section, metadata):
+    """
+    Translate template representing Mercator.
+
+    Updates the metadata in-place with the translations.
+
+    Args:
+
+    * section:
+        Dictionary of coded key/value pairs from section 3 of the message.
+
+    * metadata:
+        :class:`collections.OrderedDict` of metadata.
+
+    """
+    major, minor, radius = ellipsoid_geometry(section)
+    geog_cs = ellipsoid(section['shapeOfTheEarth'], major, minor, radius)
+
+    true_scale_lat = section['LaD'] * _GRID_ACCURACY_IN_DEGREES
+
+    cs = icoord_systems.Mercator(true_scale_lat=true_scale_lat,
+                                 ellipsoid=geog_cs)
+
+    # Create the X and Y coordinates.
+    x_coord, y_coord, scan = _calculate_proj_coords_from_lon_lat(section, cs)
+
+    # Determine the lat/lon dimensions.
+    y_dim, x_dim = 0, 1
+    if scan.j_consecutive:
+        y_dim, x_dim = 1, 0
+
+    # Add the X and Y coordinates to the metadata dim coords.
+    metadata['dim_coords_and_dims'].append((y_coord, y_dim))
+    metadata['dim_coords_and_dims'].append((x_coord, x_dim))
+
+
 def grid_definition_template_12(section, metadata):
     """
     Translate template representing transverse Mercator.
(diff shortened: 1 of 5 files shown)


iris_grib/_load_convert.py
```diff
@@ -848,17 +888,30 @@ def _calculate_proj_coords_from_lon_lat(section, cs):
...
+        raise TranslationError('Unsupported lat-lon point parameters')
+
     scan = scanning_mode(section['scanningMode'])
     lon_0 = section['longitudeOfFirstGridPoint'] * _GRID_ACCURACY_IN_DEGREES
     lat_0 = section['latitudeOfFirstGridPoint'] * _GRID_ACCURACY_IN_DEGREES
     x0_m, y0_m = cs.as_cartopy_crs().transform_point(
         lon_0, lat_0, ccrs.Geodetic())
-    dx_m = section['Dx'] * 1e-3
-    dy_m = section['Dy'] * 1e-3
+    dx_m = dx * 1e-3
```
dkillick: Are you sure this constant offset applies equally to i/j as well as to x/y?

iris_grib/_load_convert.py
```diff
@@ -1195,6 +1248,9 @@ def grid_definition_section(section, metadata):
     elif template == 5:
         # Process variable resolution rotated latitude/longitude.
         grid_definition_template_5(section, metadata)
+    elif template == 10:
+        # Process Mercator
```
dkillick: \<pedant>Full stop at the end of the comment, please.\</pendant>

iris_grib/message.py
```diff
@@ -33,6 +33,8 @@
 from iris._lazy_data import as_lazy_data
 from iris.exceptions import TranslationError
 
+SUPPORTED_GRID_DEFINITIONS = (0, 1, 5, 10, 12, 20, 30, 40, 90)
```
dkillick: What benefit did you get for factoring this out to the top of the file? I don't see it's a problem, I'm just interested...

You could probably do with making it a private global variable though (add a leading underscore `_`), as per the similar variables in `_load_convert`.

iris_grib/tests/unit/load_convert/test_grid_definition_template_10.py
```diff
@@ -0,0 +1,100 @@
+# (C) British Crown Copyright 2014 - 2018, Met Office
```
dkillick: Copyright needs to start at 2018 for this new file!

iris_grib/_load_convert.py
dkillick: You could add some testing to confirm whether it does or not 😄

kaedonkers: I thought it did but having read the template definitions more closely the units of Di/Dj/Dx/Dy can have units of 10^-2 m, 10^-3 m, 10^-6 degrees or None :worried:

iris_grib/_load_convert.py
pelson: Frankly, the word "process" is superfluous too. 😉

iris_grib/_load_convert.py
kaedonkers: I will refactor it to only use grid length (which are still either `Di`+`Dj` or `Dx`+`Dy`). These all use 10^-3 m as their unit. I have renamed things accordingly.

dkillick:
@pelson things I might have missed 😉 

> I'm too far away from GRIB

This is a thing we're also hitting, so the thinking was that many pairs of eyes are often better than just the one 👀

kaedonkers:
@dkillick I made the changes you suggested, tidied up a couple things and added this feature to the new release notes.

The tests are still failing but that is because Travis uses the Iris v2.0 release which does not yet support the `standard_parallel` kwarg for the Mercator projection (all other tests pass). This kwarg should be supported in Iris v2.1. So we will need to wait until that is released to merge this PR.

iris_grib/tests/unit/load_convert/test_grid_definition_template_10.py
```diff
@@ -0,0 +1,100 @@
...
+                                 coord_system=cs)
+        expected['dim_coords_and_dims'].append((y, y_dim))
+        expected['dim_coords_and_dims'].append((x, x_dim))
+        return expected
+
+    def test(self):
+        section = self.section_3()
+        metadata = empty_metadata()
+        grid_definition_template_10(section, metadata)
+        expected = self.expected(0, 1)
```
dkillick: It's not very clear what 0 and 1 mean here. Perhaps you could clarify them?

iris_grib/_load_convert.py
```diff
@@ -844,22 +888,37 @@ def grid_definition_template_20(section, metadata):
     metadata['dim_coords_and_dims'].append((x_coord, x_dim))
 
 
-def _calculate_proj_coords_from_lon_lat(section, cs):
+def _calculate_proj_coords_from_grid_lengths(section, cs):
     # Construct the coordinate points, the start point is given in millidegrees
     # but the distance measurement is in 10-3 m, so a conversion is necessary
     # to find the origin in m.
 
+    MM_TO_M = 1e-3
```
dkillick: It's a small thing, but by convention a variable named in all-caps is used to indicate a global variable. As this is defined in a non-global context, perhaps you could make this variable name lower-case.

iris_grib/_load_convert.py
```diff
@@ -844,22 +888,37 @@ def grid_definition_template_20(section, metadata):
...
+
     scan = scanning_mode(section['scanningMode'])
     lon_0 = section['longitudeOfFirstGridPoint'] * _GRID_ACCURACY_IN_DEGREES
     lat_0 = section['latitudeOfFirstGridPoint'] * _GRID_ACCURACY_IN_DEGREES
     x0_m, y0_m = cs.as_cartopy_crs().transform_point(
         lon_0, lat_0, ccrs.Geodetic())
-    dx_m = section['Dx'] * 1e-3
-    dy_m = section['Dy'] * 1e-3
+    dx_m = dx * MM_TO_M
+    dy_m = dy * MM_TO_M
```
dkillick: You'll then need to update the variable name on these two lines as well.

Issue #14, opened 2016-06-02
Port SciTools/iris#2020


bjlittle:
https://github.com/SciTools/iris/pull/2020

PR #107, opened 2018-06-06
Add a release-notes section.

pp-mo:
I'm feeling the need for this, since #106 is a bugfix with behaviour change.

Coming up soon : 
  * further fix to Hybrid Pressure loading to get an actual factory
  * saving for both hybrid height and hybrid pressure

... so I think change notes will be useful

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -91,6 +91,29 @@ using ``iris-grib``::
     >>> iris.save(my_cube, 'my_file.grib2')
 
 
+Getting Started
+===============
+
+To ensure all ``iris-grib`` dependencies, it is sufficient to have installed
+:mod:`Iris <iris>` itself.
+
+The simplest way to install is with
+`conda <https://conda.io/miniconda.html>`_ ,
+using the `conda-forge channel <https://anaconda.org/conda-forge>`_ ,
+with the command
+
+    $ conda install iris-grib -c conda-forge
+
+Development sources are hosted at `<https://github.com/SciTools/iris-grib>`_ .
+
+Releases
+--------
+
+The latest release is v0.12.x .
+
+For recent changes, see `Release Notes <ref/release_notes.html>`_ .
+
+
 Indices and tables
 ==================
 
--- a/docs/ref/release_notes.rst
+++ b/docs/ref/release_notes.rst
@@ -1,0 +1,54 @@
+What's new in Iris-grib v0.13
+=============================
+
+:Release: 0.13.0
+:Date: T.B.D.
+
+Features
+--------
+
+* Added loading for data on Hybrid Height levels (surface type 118 in
+  code table 4.5).
+
+Bug Fixes
+---------
+
+* Fixed a bug with loading data on Hybrid Pressure levels (surface types 105
+  and 119 in code table 4.5).  
+  Previously, *all* hybrid coordinate values, in both 'level_pressure' and
+  'sigma' coordinates, were loaded from the next level up,
+  i.e. (model_level_number + 1).
+
+  .. note::
+
+      This changes loading behaviour : previous results were simply wrong.
+
+
+What's new in Iris-grib v0.12
+=============================
+
+:Release: 0.12
+:Date: 25 Oct 2017
+
+Use `ecCodes <https://software.ecmwf.int/wiki/display/ECC/ecCodes+Home>`_ for
+the file interface.
+This is ECMWF's replacement for the older GRIB-API, which is now deprecated.
+
+
+What's new in Iris-grib v0.11
+=============================
+
+:Release: 0.11
+:Date: 25 Oct 2017
+
+Update for Iris v2.0+, using Dask instead of Biggus
+
+
+What's new in Iris-grib v0.9
+=============================
+
+:Release: 0.9.0
+:Date: 25 Jul 2016
+
+Stable release of iris_grib to support iris v1.10
+


pp-mo:
N.B. to test, 
  * you (just) need an env with : iris, eccodes and sphinx
  * then `cd docs; make html`

We could document that, I suppose ... ? 
( we don't bother at present for Iris, but we do now have automated docs-building there, so you can work it out from that )

docs/index.rst
```diff
@@ -91,6 +91,30 @@ using ``iris-grib``::
...
+To ensure all ``iris-grib`` dependencies, it is sufficient to have installed
+:mod:`Iris <iris>` itself, and
+`ecCodes <https://software.ecmwf.int/wiki/display/ECC/ecCodes+Home>`_ .
+
+The simplest way to install is with
+`conda <https://conda.io/miniconda.html>`_ ,
+using the `conda-forge channel <https://anaconda.org/conda-forge>`_ ,
+with the command
+
+    $ conda install iris-grib -c conda-forge
```
corinnebosley: I'm still not quite sure what our position is with respect to advising people to use conda.  I'll have a chat with Pete about this I suppose.

docs/ref/release_notes.rst
```diff
@@ -0,0 +1,54 @@
...
+
+* Fixed a bug with loading data on Hybrid Pressure levels (surface types 105
+  and 119 in code table 4.5).  
+  Previously, *all* hybrid coordinate values, in both 'level_pressure' and
+  'sigma' coordinates, were loaded from the next level up,
+  i.e. (model_level_number + 1).
+
+  .. note::
+
+      This changes loading behaviour : previous results were simply wrong.
```
corinnebosley: Maybe we should use a slightly different way of wording this than 'simply wrong' (which is true, but a little blunt).  How about 'incorrect'?

docs/index.rst
pp-mo: > have a chat with Pete

you mean @dkillick ;-) 
Good idea. 
Well, Iris docs has gone over to pip, but its README still recommends conda. 
I certainly still think conda is most reliable + I wouldn't use anything else for choice myself.
IMO this ought to be fine, as (a) it's for external users and (b) we don't provide a pip install for iris-grib anyway (though perhaps we should, if that's the way it's going).

corinnebosley: Yeah sorry, I will have a chat with @dkillick because we are virtual people now.

docs/ref/release_notes.rst
pp-mo: The point I'm trying to make is that, although this a breaking behaviour change, we can allow that because the existing behaviour is definitely "wrong", i.e. you can't really defend what it currently does as being of use to anyone (!)
I'll try + come up with something milder ..

docs/index.rst
dkillick: I actually don't exist.

P.S. I think saying "use conda" is fine, and we [recommend it for Iris](https://scitools.org.uk/iris/docs/latest/installing.html#installing-using-conda).

dkillick: > we don't provide a pip install for iris-grib ... (though perhaps we should)

I think we should. Many other SciTools packages are now on PyPI, and we probably should support it as PyPI remains popular.

docs/ref/release_notes.rst
```diff
@@ -0,0 +1,54 @@
...
+=============================
+
+:Release: 0.13.0
+:Date: T.B.D.
+
+Features
+--------
+
+* Added loading for data on Hybrid Height levels (surface type 118 in
+  code table 4.5).
```
dkillick: Assumedly we'll add GDT10 load/save support to this list as and when it arrives..?

docs/ref/release_notes.rst
dkillick: @pp-mo I support the suggestion of @corinnebosley that we could do with some milder wording for this!

We could also make it absolutely clear that the changed load behaviour affects _only_ loading hybrid pressure levels.

docs/ref/release_notes.rst
```diff
@@ -0,0 +1,54 @@
+What's new in Iris-grib v0.13
```
dkillick: We should have a consistent way of rendering `iris-grib` everywhere. My preference is for all lower-case and hyphen-separated (i.e. "iris-grib"), as that's how it's referred to elsewhere, notably in the README.

docs/ref/release_notes.rst
```diff
@@ -0,0 +1,54 @@
...
+
+
+What's new in Iris-grib v0.12
+=============================
+
+:Release: 0.12
+:Date: 25 Oct 2017
+
+Use `ecCodes <https://software.ecmwf.int/wiki/display/ECC/ecCodes+Home>`_ for
+the file interface.
```
dkillick: I wonder if "the file interface" is a little obscure. Perhaps

    ... as the interface to GRIB files...

would be clearer?

docs/ref/release_notes.rst
```diff
@@ -0,0 +1,54 @@
...
+Update for Iris v2.0+, using Dask instead of Biggus
+
+
+What's new in Iris-grib v0.9
+=============================
+
+:Release: 0.9.0
+:Date: 25 Jul 2016
+
+Stable release of iris_grib to support iris v1.10
```
dkillick: Ooh, here's another different variant of "iris-grib" :wink:

docs/ref/release_notes.rst
```diff
@@ -0,0 +1,54 @@
...
+This is ECMWF's replacement for the older GRIB-API, which is now deprecated.
+
+
+What's new in Iris-grib v0.11
+=============================
+
+:Release: 0.11
+:Date: 25 Oct 2017
+
+Update for Iris v2.0+, using Dask instead of Biggus
```
dkillick: We could add doc links for dask (and optionally biggus)?

docs/ref/release_notes.rst
```diff
@@ -0,0 +1,54 @@
+What's new in Iris-grib v0.13
+=============================
+
+:Release: 0.13.0
+:Date: T.B.D.
```
dkillick: I take it these (that is, `:Release:` and `:Date:`) have defined functionality in ReStructured Text?

docs/index.rst
pp-mo: Ok, but I don't think it needs to be done right now.
#14

docs/ref/release_notes.rst
pp-mo: Absolutely, it should be incremental.
And hopefully, we'll all get in the habit of keeping it up-to-date.
I don't think we need an Iris-style |'contributions + aggregate' process.

docs/ref/release_notes.rst
pp-mo: :+1:

docs/index.rst
dkillick: > I don't think it needs to be done right now

Agreed. Thanks for raising an issue about it.

docs/ref/release_notes.rst
pp-mo: TBH I have no idea. 
I don't think they do anything special -- I just copied the cartopy one.

docs/ref/release_notes.rst
pp-mo: :+1:

docs/ref/release_notes.rst
dkillick: Let's just go with it!

pp-mo:
Thanks for input @corinnebosley @dkillick 
See what you think now ?

corinnebosley:
@pp-mo I'm happy with this, looks good.  I'll wait for collation of approvals, but I'll make sure it gets merged today.

kaedonkers:
@corinnebosley Oops, sorry I was going to review this morning. Looks good to me though :+1:

I'll rebase #104 and add "GDT10 load support" to the release notes.

PR #113, opened 2018-06-12
Save Grid Definition Template 4 

dkillick:
Adds functionality to save cubes as GRIB2 that have an irregular, non-rotated, lat/lon grid.


iris_grib/_save_rules.py
```diff
@@ -576,9 +591,7 @@ def grid_definition_section(cube, grib):
 
     if isinstance(cs, GeogCS):
         if not regular_x_and_y:
-            raise iris.exceptions.TranslationError(
-                'Saving an irregular latlon grid to GRIB (PDT3.4) is not '
-                'yet supported.')
+            grid_definition_template_4(cube, grib)
```
pp-mo: Surely there should be an "else" here ??
Otherwise it calls `grid_definition_template_4` **and then** `grid_definition_template_0`
Which sounds wrong ...

dkillick: Yup, I think there should 😳 Fixing this should get that test passing too 👍

dkillick:
I've added a new integration test class (and module) that checks the correct grid definition section is chosen to save the horizontal grid. Given the earlier test failure, it seems that such a test is a good idea...

iris_grib/_save_rules.py
```diff
@@ -611,8 +611,9 @@ def grid_definition_section(cube, grib):
         # Lambert Conformal coordinate system (template 3.30).
         grid_definition_template_30(cube, grib)
     else:
-        raise ValueError('Grib saving is not supported for coordinate system: '
-                         '{}'.format(cs))
+        name = cs.grid_mapping_name.replace('_', ' ').title()
```
pp-mo: I don't think the `.title()` should be there : `cs.grid_mapping_name` is just a string.
Presumably you tested this, hence spotting the problem?
Maybe it even needs an actual test.

iris_grib/tests/integration/save_rules/test_grid_definition_section.py
```diff
@@ -0,0 +1,129 @@
...
+    def test_coord_system_not_supported(self):
+        # Test an unsupported grid - let's choose Albers Equal Area.
+        x_points = np.arange(3)
+        y_points = np.arange(3)
+        cs = AlbersEqualArea(ellipsoid=self.ellipsoid)
+        test_cube = self._make_cube(x_points, y_points, cs)
+
+        exp_name = cs.grid_mapping_name.replace('_', ' ').title()
+        exp_emsg = 'not supported for coordinate system {!r}'.format(exp_name)
+        with self.assertRaisesRegexp(ValueError, exp_emsg):
```
pp-mo: I assume from the above that there is an error, which does contain the expected string ... but it is not actually the error that we were expecting !!

iris_grib/tests/integration/save_rules/test_grid_definition_section.py
```diff
@@ -0,0 +1,129 @@
...
+from iris_grib._save_rules import grid_definition_section
+from iris_grib.tests.unit.save_rules import GdtTestMixin
+
+
+class Test(tests.IrisGribTest, GdtTestMixin):
+    def setUp(self):
+        GdtTestMixin.setUp(self)
+        self.ellipsoid = GeogCS(6371200)
+
+    def _make_cube(self, x_points, y_points, cs, coord_units='1'):
```
pp-mo: Could you not extend `GdtTestMixin._make_test_cube` to do this, instead of making a new thing ?
It looks like you want to duplicate that, but take control of the x/y names and units, so you could just add keywords for that.
(It already has coord-system control)

iris_grib/_save_rules.py
dkillick: @pp-mo I'm just making the `grid_mapping_name` more human-readable for the sake of the error message, that is:

`lambert_conformal` --> `Lambert Conformal`

pp-mo: Sure, but is the method really general?  It depends on the content of the strings.
I think staying closer to the code (the name of the thing) might be actually clearer here.

pp-mo: Oh, 'title' is a standard string method  :mortar_board: 
Ok then, I just didn't get that this would actually work !!

iris_grib/tests/unit/save_rules/test_grid_definition_template_4.py
```diff
@@ -0,0 +1,99 @@
...
+    def test__grid_shape(self):
+        test_cube = self._make_test_cube(x_points=np.arange(13),
+                                         y_points=np.arange(6))
+        grid_definition_template_4(test_cube, self.mock_grib)
+        self._check_key('Ni', 13)
+        self._check_key('Nj', 6)
+
+    def test__grid_points(self):
+        x_floats = np.array([11.0, 12.0, 167.0])
+        # TODO: reduce Y to 2 points, when gribapi nx=ny limitation is gone.
```
corinnebosley: Wow, is that really a thing?  Who thought that would be a good idea?!

iris_grib/tests/integration/save_rules/test_grid_definition_section.py
dkillick: @pp-mo yup - I had to add exactly this functionality in #116 anyhow, so I'll just make use of it here too.

iris_grib/tests/unit/save_rules/test_grid_definition_template_4.py
dkillick: The TODO says it is (or it was)... I haven't investigated it further as I don't think it impacts these tests at all.

pp-mo: > I haven't investigated it further as I don't think it impacts these tests

Ooh :+1: for work-not-done ! :mortar_board:

dkillick:
Remove `tests/unit/save_rules/test_grid_definition_section` as one of the two tests is obsoleted by the code additions here and the other duplicates testing performed in the new and more appropriately located integration tests for `_save_rules.grid_definition_section`.

PR #114, opened 2018-06-12
Hybrid coords round trip

corinnebosley:
Round-trip integration tests for hybrid height and hybrid pressure.

Also a small fix for a bit that was forgotten and only revealed in these tests.


corinnebosley:
@dkillick @pp-mo Should I put in some bits at the end of each test to delete the saved files?

pp-mo:
"Otherwise" LGTM :+1:  -- short, but if it works that's already proving quite a lot.
Sorry, meant to create a proper review but pressed wrong button.

corinnebosley:
@pp-mo Thanks for that, useful pointers, especially the temp_filepath pattern that I didn't know about before.  Cool.

corinnebosley:
Wait a minute though, I still need to delete the save directories that I had before...

corinnebosley:
@pp-mo @dkillick @kaedonkers Tests now passing.  Any more tweaks?

pp-mo:
> Any more tweaks?

Maybe just a comment to say that/why it is loading+save a separate reference cube.

I'm also interested in your `save(... saver='grib2')` usage as I didn't know it could do that ! 
The docstrings as-is suggest it can only be a saver *function*.
Obviously that's is an Iris issue...

corinnebosley:
@pp-mo Yeah I didn't know you could do that either, it was a bit of guesswork that got lucky because the docs are really really vague.

corinnebosley:
@pp-mo Thanks!

PR #116, opened 2018-06-12
Save Mercator Grid Definition Template

dkillick:
Adds functionality to save cubes on the Mercator projection to GRIB2, grid definition template 10.

Thanks to @kaedonkers for doing most of the work represented here!

### DIFF
--- a/iris_grib/_save_rules.py
+++ b/iris_grib/_save_rules.py
@@ -523,7 +523,7 @@ def grid_definition_template_12(cube, grib):
         x_step = step(x_cm, atol=1)
         y_step = step(y_cm, atol=1)
     except ValueError:
-        msg = ('Irregular coordinates not supported for transverse '
+        msg = ('Irregular coordinates not supported for Transverse '
                'Mercator.')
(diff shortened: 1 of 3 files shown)


iris_grib/_save_rules.py
```diff
@@ -455,7 +526,7 @@ def grid_definition_template_12(cube, grib):
         x_step = step(x_cm, atol=1)
         y_step = step(y_cm, atol=1)
     except ValueError:
-        msg = ('Irregular coordinates not supported for transverse '
+        msg = ('Irregular coordinates not supported for Transverse '
```
corinnebosley: I'm in two minds about whether this should have the word 'yet' between 'not' and 'supported'.  It sounds more accommodating, but does also imply that we intend to do it some day, and I would imagine that if that is the case, it's a very long way off...

Maybe let's just leave it as it is.

dkillick: @corinnebosley it might be that an irregular Transverse Mercator grid is not supported by GRIB itself.

PR #121, opened 2018-06-13
WIP: Steal iris tests

corinnebosley:
Lifted and shifted iris tests which are for iris-grib and hence no longer belong in iris.

Some changes necessary, especially to remove any unnecessary or unreasonable mocking.

### DIFF
--- a/iris_grib/tests/unit/grib_phenom_translation/test_timecodes.py
+++ b/iris_grib/tests/unit/grib_phenom_translation/test_timecodes.py
@@ -1,0 +1,269 @@
+# (C) British Crown Copyright 2014 - 2016, Met Office
+#
+# This file is part of iris-grib.
+#
+# iris-grib is free software: you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# iris-grib is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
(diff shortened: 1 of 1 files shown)


corinnebosley:
Tests are not yet passing due to an InternalGribError which occurs in a type-check of the grib-message dictionary keys.  For some reason a wrapper decorator puts a dictionary inside a tuple, so when the function later checks that the values are integers, it actually finds that the one and only value is an entire dictionary.

What bothers me is that this was not the case when this test was run in iris, so what's different?

corinnebosley:
I know what's different.  The iris test mocks out any use of gribapi, so it's never actually tested.  As soon as I remove the mock-up after copying into iris-grib testing, the test fails.

I HATE MOCK....

iris_grib/tests/unit/grib_phenom_translation/test_timecodes.py
```diff
@@ -0,0 +1,274 @@
...
+
+
+class TestGribTimecodes(tests.IrisGribTest):
+    def _run_timetests(self, test_set):
+        # Check the unit-handling for given units-codes and editions.
+
+        # Operates on lists of cases for various time-units and grib-editions.
+        # Format: (edition, code, expected-exception,
+        #          equivalent-seconds, description-string)
+        for test_controls in test_set:
```
corinnebosley: Just above this line, this iris test had this: https://github.com/SciTools/iris/blob/master/lib/iris/tests/test_grib_load_translations.py#L206

That allowed the test to work because it avoided the gribapi type-check.  Without that line, gribapi messes everything up.

bjlittle:
@corinnebosley Where are you with this PR? Is it still alive?

corinnebosley:
@bjlittle I think this died a long time ago.  I think this was part of the work we did in separating iris and iris-grib, and I think we either decided that this wasn't necessary or @DPeterK overrode it.  I'm closing it.

PR #122, opened 2018-06-13
Save GDT20 - Polar Stereographic

DPeterK:
Add save support for Grid Definition Template 20 - Polar Stereographic.

### DIFF
--- a/iris_grib/_save_rules.py
+++ b/iris_grib/_save_rules.py
@@ -640,13 +640,6 @@ def _perspective_projection_common(cube, grib):
     gribapi.grib_set(grib, "LaD", cs.central_lat / _DEFAULT_DEGREES_UNITS)
     gribapi.grib_set(grib, "LoV", central_lon / _DEFAULT_DEGREES_UNITS)
 
-    # Which pole are the parallels closest to? That is the direction
-    # that the cone converges.
-    latin1, latin2 = cs.secant_latitudes
-    poliest_secant = latin1 if abs(latin1) > abs(latin2) else latin2
-    centre_flag = 0x0 if poliest_secant > 0 else 0x1
-    gribapi.grib_set(grib, 'projectionCentreFlag', centre_flag)
-
 
 def grid_definition_template_20(cube, grib):
     """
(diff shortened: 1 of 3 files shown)


iris_grib/tests/integration/save_rules/test_grid_definition_section.py
```diff
@@ -27,14 +27,13 @@
...
 
-from iris.coords import DimCoord
 from iris.coord_systems import (GeogCS,
                                 RotatedGeogCS,
                                 Mercator,
                                 TransverseMercator,
+                                Stereographic,
                                 LambertConformal,
                                 AlbersEqualArea)
-from iris.cube import Cube
```
dkillick: Unused imports...

iris_grib/_save_rules.py
```diff
@@ -634,17 +633,63 @@ def grid_definition_template_30(cube, grib):
...
+
+    # Retrieve the cube coord system.
+    cs = cube.coord(dimensions=[0]).coord_system
+
+    _perspective_projection_common(cube, grib)
+
+    # Is this a north or south polar stereographic projection?
+    if cs.central_lat == -90.0:
+        # XXX: is this the correct value? We read bit 0x80 in load, but set bit
+        #      0x1 when saving GDT30...
```
corinnebosley: I don't know, is it the correct value? This is a very unnerving comment.

pp-mo:
I'm not happy with this as-is, 'cos I think we have some outstanding problems + we want some more real(er) testing.

Given the bug uncovered [above](https://github.com/SciTools/iris-grib/pull/122#discussion_r195332221), I'm thinking that at least one of NorthPolar/SouthPolar would currently fail a save-then-load roundtrip check. 
That's a good reason for having one.

iris_grib/_save_rules.py
dkillick: > This is a very unnerving comment

Yup, it is. I assumed that it should be `0x80` but it was a bit odd that the GDT30 saver did not set that value...

pp-mo: > it was ***a bit odd*** that the GDT30 saver did not set that value

Bugs in existing implementation -- doesn't sound that unusual to me :rofl:

bjlittle:
@DPeterK Fancy re-basing? :smile:

DPeterK:
@bjlittle on it!

bjlittle:
@pp-mo Do you fancy taking this PR on, and getting across the finish line? 
:car: :checkered_flag:

PR #123, opened 2018-06-14
Soft deps in setup

corinnebosley:
Popped a little function in the setup.py to look for iris 2.1 because that would be best, but if it's not available then it will install using iris 2.0 (or later).

Also added the eccodes dependency.  gribapi lives in python-eccodes, but I guess the (official?) package name is actually just eccodes.

### DIFF
--- a/iris_grib/_load_convert.py
+++ b/iris_grib/_load_convert.py
@@ -43,6 +43,7 @@ from iris.fileformats.rules import ConversionMetadata, Factory, Reference, \
     ReferenceTarget
 from iris.util import _is_circular
 
+from . import _confirm_iris_mercator_support
 from ._grib1_load_rules import grib1_convert
 from .message import GribMessage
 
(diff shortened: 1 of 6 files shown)


corinnebosley:
@pp-mo This is a really nice solution.  Nice one.

iris_grib/_load_convert.py
```diff
@@ -43,6 +43,7 @@
     ReferenceTarget
 from iris.util import _is_circular
 
+from . import _confirm_iris_mercator_support
```
pp-mo: Sorry I this is bugged.
I think it should really be `from ..  import ` ??
Always a bit confused by relative imports ! :confused:

pp-mo:
> confused by relative imports

Yikes, I created a circular import problem !
https://github.com/corinnebosley/iris-grib/pull/2 should fix this, I hope...

iris_grib/_load_convert.py
corinnebosley: I think it works though, but now it can't find `grib_phenom_translation`.  Weird...

corinnebosley:
@kaedonkers Ping!

corinnebosley:
@bjlittle Ping

pp-mo:
> @kaedonkers @bjlittle Ping!

Seconded -- can someone who is not me take a look please ?!?

PR #135, opened 2018-10-23
added explicit flag to conda env list

corinnebosley:
- Added `--explicit` to conda env list so that we can grab the environment manifest of each run to reproduce it.
- Added `conda info -a` to list all info.  Not committed to this addition, might be worthwhile though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,8 @@ install:
   # Install iris-grib's dependencies
   # --------------------------------
   - conda install --quiet --yes python=${python} iris python-eccodes numpy cartopy mock filelock pep8;
-  - conda list
+  - conda list --explicit;
+  - conda info -a;
 
   # Download iris-test-data
   # --------------------------------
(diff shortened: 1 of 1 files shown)


corinnebosley:
Ping @dkillick !

.travis.yml
```diff
@@ -46,7 +46,7 @@ install:
 
   # Summarise the environment
   # -------------------------
-  - conda list
+  - conda list --explicit
```
dkillick: @corinnebosley I think you need to explicitly list the environment **before** `iris-grib` is added to it. This way you can reconstruct the environment _into which_ `iris-grib` may be installed. If you list it after `iris-grib` has been (pip-)installed, you might end up listing an environment that can't be automatically reconstructed. 

Perhaps try adding `--explicit` back to [L20](https://github.com/SciTools/iris-grib/blob/f954435fbaec39a1ed1ed493adde6611f217a4d1/.travis.yml#L20) and reverting the change here:

```suggestion
  - conda list
```

corinnebosley:
Ping @dkillick

PR #137, opened 2018-11-23
Product Definition Template 60

corinnebosley:
Translations for "Individual ensemble reforecast, control and perturbed, at a horizontal level or in a horizontal layer at a point in time."


bjlittle:
@corinnebosley What's the status of this PR? Still active?

If so, fancy resolving the testing issues?

corinnebosley:
@bjlittle No, this is what @pp-mo and I are working on trying to scope out still.  This isn't just a quick fix, it's gonna take a lot of work and I'm tempted to close this and start again because I think it will be cleaner.

PR #138, opened 2019-01-03
Get iris-grib working in Python 3 !

lbdreyer:
The eccodes 2.10 release includes Python 3 support.

This PR includes Python 3 in the testing matrix and thus highlights outstanding issues with iris-grib/eccodes/Python3

Known issues with Python 3 that work fine in Python 2:
- [x] **1) Can't set the key 'NV' as np.int64**
   **TESTS:** integration.save_rules.test_save_hybrid_coords (all tests in file)
- [x] **2) Problem with lazy loading of _data_ from files**
    **2.i) files with hybrid factory - problem loading orography field during factory creation**
   files: iris_grib/tests/testdata/faked_sample_hh_grib_data.grib2 and faked_sample_hp_grib_data.grib2
   **TESTS:** integration.load_convert.test_load_hybrid_coords (all tests in file) and integration.round_trip.test_hybrid_coords_round_trip (all tests in file)
    **2.ii)) Can load in file as cube, when trying to call the data it fails**
  files: iris_grib/tests/testdata/*.grib2, (excet CB which does work)
   **TESTS:** integration.round_trip/test_WAFC_mapping_round_trip all tests in file except CB.grib2)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,20 +17,18 @@ install:
 
   # Install iris-grib's dependencies
   # --------------------------------
-  # If running in Python 2, install Python2-only python-eccodes.
-  - if [[ "$python" == "2.7" ]]; then
+  # If running in Python 2, install Python2-only python-eccodes, otherwise we
+  # install the Python3-only eccodes-python via pip.
+  - if [[ "${python}" == "2.7" ]]; then
         export ECCODES_DEPS="python-eccodes";
-    fi
-
-  # If running in Python 3, we will install Python3-only eccodes-python via pip.
-  - if [[ "$python" == "3.6" ]]; then
+    else
         export ECCODES_DEPS="eccodes pip";
     fi
 
-  - conda install --quiet --yes python=${python} iris numpy cartopy mock filelock pep8 $ECCODES_DEPS;
+  - conda install --quiet --yes python=${python} iris numpy cartopy mock filelock pep8 ${ECCODES_DEPS};
     
   # If running in Python 3, install eccodes-python and check the installation.
-  - if [[ "$python" == "3.6" ]]; then
+  - if [[ "${python}" != "2.7" ]]; then
         pip install eccodes-python;
         python -m eccodes selfcheck;
     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: python
 sudo: false
 env:
   - python=2.7
+  - python=3.6
 
 install:
   # Fetch and install conda
@@ -16,7 +17,24 @@ install:
 
   # Install iris-grib's dependencies
   # --------------------------------
-  - conda install --quiet --yes python=${python} iris python-eccodes numpy cartopy mock filelock pep8;
+  # If running in Python 2, install Python2-only python-eccodes.
+  - if [[ "$python" == "2.7" ]]; then
+        export ECCODES_DEPS="python-eccodes";
+    fi
+
+  # If running in Python 3, we will install Python3-only eccodes-python via pip.
+  - if [[ "$python" == "3.6" ]]; then
+        export ECCODES_DEPS="eccodes pip";
+    fi
+
+  - conda install --quiet --yes python=${python} iris numpy cartopy mock filelock pep8 $ECCODES_DEPS;
+    
+  # If running in Python 3, install eccodes-python and check the installation.
+  - if [[ "$python" == "3.6" ]]; then
+        pip install eccodes-python;
+        python -m eccodes selfcheck;
+    fi
+
   - conda list --explicit
 
   # Download iris-test-data
@@ -29,7 +47,7 @@ install:
 
   # Locate iris installation
   # ------------------------
-  - export IRIS_DIR=$(python -c "import iris; import os.path; print os.path.dirname(iris.__file__)")
+  - export IRIS_DIR=$(python -c "import iris; import os.path; print(os.path.dirname(iris.__file__))")
 
   # Poke site.cfg to reference iris-test-data
   - SITE_CFG=$IRIS_DIR/etc/site.cfg
--- a/iris_grib/_save_rules.py
+++ b/iris_grib/_save_rules.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2018, Met Office
+# (C) British Crown Copyright 2010 - 2019, Met Office
 #
 # This file is part of iris-grib.
 #
@@ -976,7 +976,8 @@ def set_fixed_surfaces(cube, grib, full3d_cube=None):
             coeffs_array[n_lev - 1] = height
             coeffs_array[n_coeffs + n_lev - 1] = sigma
         pv_values = [float(el) for el in coeffs_array]
-        gribapi.grib_set(grib, "NV", n_coeffs * 2)
+        # eccodes does not support writing numpy.it64, cast to python int
+        gribapi.grib_set(grib, "NV", int(n_coeffs * 2))
         gribapi.grib_set_array(grib, "pv", pv_values)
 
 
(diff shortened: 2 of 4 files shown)


rsignell-usgs:
This is our best bet to get GRIB-writing going with Python 3, right?

ocefpaf:
> This is our best bet to get GRIB-writing going with Python 3, right?

The alternative is the "raw" python-eccodes, wrapped here, but I do not recommend it. `iris-grib` is much better and provide the high level interface we need.

@bjlittle and @lbdreyer are the errors here to hard to fix? I'm not sure what to make of all those 'grib message errors' there :-/

bjlittle:
@rsignell-usgs and @ocefpaf Unfortunately, we can't work around (in iris-grib) the issues that we're hitting due to the problems with ecCodes in Python3. As a result, we've not deployed GRIB support for Python3 here at the Met Office. That's a major blocker. We've escalated this at an organisational level with the ECMWF. As a result, they are scheduled to come to us after Easter to discuss this further and update us on their progress with estimates of when ecCodes will be fully functional in Python3.

HTH

ocefpaf:
Thanks for the clarifications @bjlittle! Let me know if I can do anything to help.

lbdreyer:
I have changed this PR so that it uses eccodes-python for Python 3 (as eccodes-python is Python 3 only) and uses python-eccodes for Python 2.

It looks like eccodes-python has fixed most of our problems with Python 3. :tada: :tada: :tada:

However, you will notice that to get the tests passing I had to include some unpleasant workarounds. I am definitely not suggesting we use these workarounds, I just wanted to show that it would work. I am going to raise the outstanding issues with the developers of eccodes-python, namely that:
* Sometimes grib returns an array when we expect a list (e.g. `gribapi.grib_get_double_array' returns a list)
* It fails if we try to set the`NV` key with a object of type `numpy.int64` and so we have to convert the Python with `int()`

lbdreyer:
I have updated this PR with a better workaround for when `gribapi.get_*array` returns a list as opposed to a numpy array.

This PR now contains two commits:
commit 1: introduces testing with Python3/eccodes-python; adds a few fixes necessary for the latest eccodes.
commit 2: includes the *temporary* wrapper that we can use whenever we call `gribapi.grib_get_*array`
I have split it like this as my intention is to remove the contents of commit 2 once eccodes-python has been fixed (I have raised this issue with the developers of eccodes-python and they are looking into it)

I suggest we merge this in so that we can get users testing with iris-grib/Python3 and potentially find issues our own testing of eccodes-python has not.

lbdreyer:
@bjlittle are you able to look at this?

.travis.yml
```diff
@@ -2,6 +2,7 @@ language: python
 sudo: false
 env:
   - python=2.7
+  - python=3.6
```
bjlittle: @lbdreyer Awesome!

.travis.yml
```diff
@@ -16,7 +17,24 @@ install:
...
-  - conda install --quiet --yes python=${python} iris python-eccodes numpy cartopy mock filelock pep8;
+  # If running in Python 2, install Python2-only python-eccodes.
+  - if [[ "$python" == "2.7" ]]; then
+        export ECCODES_DEPS="python-eccodes";
+    fi
+
+  # If running in Python 3, we will install Python3-only eccodes-python via pip.
+  - if [[ "$python" == "3.6" ]]; then
+        export ECCODES_DEPS="eccodes pip";
+    fi
```
bjlittle: @lbdreyer To simply we could make this the `else` of `"2.7"`. It's one or the other.... but very soon just Python3 :smile:

.travis.yml
```diff
@@ -16,7 +17,24 @@ install:
...
+
+  - conda install --quiet --yes python=${python} iris numpy cartopy mock filelock pep8 $ECCODES_DEPS;
```
bjlittle: @lbdreyer Me being a shell snob - but could you make this `${ECCODES_DEPS}` please? :wink:

.travis.yml
```diff
@@ -16,7 +17,24 @@ install:
...
+    
+  # If running in Python 3, install eccodes-python and check the installation.
+  - if [[ "$python" == "3.6" ]]; then
+        pip install eccodes-python;
+        python -m eccodes selfcheck;
```
bjlittle: @lbdreyer Nice one :+1:

.travis.yml
```diff
@@ -16,7 +17,24 @@ install:
...
+
+  # If running in Python 3, we will install Python3-only eccodes-python via pip.
+  - if [[ "$python" == "3.6" ]]; then
+        export ECCODES_DEPS="eccodes pip";
+    fi
+
+  - conda install --quiet --yes python=${python} iris numpy cartopy mock filelock pep8 $ECCODES_DEPS;
+    
+  # If running in Python 3, install eccodes-python and check the installation.
+  - if [[ "$python" == "3.6" ]]; then
```
bjlittle: @lbdreyer This could make this `"${python}" != "2.7"` instead, which is slightly more future proof and less tied to a specific minor version of Python 3

iris_grib/_save_rules.py
```diff
@@ -976,7 +976,8 @@ def set_fixed_surfaces(cube, grib, full3d_cube=None):
             coeffs_array[n_lev - 1] = height
             coeffs_array[n_coeffs + n_lev - 1] = sigma
         pv_values = [float(el) for el in coeffs_array]
-        gribapi.grib_set(grib, "NV", n_coeffs * 2)
+        # eccodes does not support writing numpy.it64, cast to python int
```
bjlittle: @lbdreyer Change `numpy.it64` -> `numpy.int64`

iris_grib/message.py
```diff
@@ -33,9 +33,24 @@
...
+    Converts what f returns into an array.
+
+    When using eccodes-python, gribapi.grib_get_*array returns a list rather
+    than an array as expected.
+
+    This is intended as a *temporary* fix that wil be moved once eccodes-python
+    is fixed.
+
+    """
+    return np.array(f)
```
bjlittle: @lbdreyer There a danger here that by only casting to `np.array` regardlessly we could lose a masked array that's passed in.

Should be be more selective?

iris_grib/message.py
```diff
@@ -33,9 +33,24 @@
 from iris._lazy_data import as_lazy_data
 from iris.exceptions import TranslationError
 
+
 _SUPPORTED_GRID_DEFINITIONS = (0, 1, 5, 10, 12, 20, 30, 40, 90)
 
 
+def grib_get_array_wrapper(f):
```
bjlittle: @lbdreyer I guess we should really make this a private function i.e. `_grib_get_array_wrapper`.

iris_grib/message.py
lbdreyer: You're right. We could either check its an array before casting, or perhaps use `np.asanyarray` which allows masked arrays to keep the mask:
```
>>> a  = ma.array([1,2,3], mask=[0,1,1])
>>> a
masked_array(data = [1 -- --],
             mask = [False  True  True],
       fill_value = 999999)

>>> f = np.asanyarray(a)
>>> f
masked_array(data = [1 -- --],
             mask = [False  True  True],
       fill_value = 999999)
```

lbdreyer:
@bjlittle I've addressed your review actions. Do you mind taking another look?

bjlittle:
@lbdreyer Let's target this as a `0.14.1` release.

After merging this PR I created a `0.14.x` branch. Previously, I stuffed up and created a `0.14.x` branch for the `0.14.0` release but forget to update the docs (doh!), rather I done that on `master`.

Anyways, please use the new `0.14.x` branch for the `0.14.1` tag and release (which already includes this PR), and update the docs on `0.14.x` (although there may be issues building them on readthedocs due to eccodes not being available from PyPI - but don't let that block the release)

bjlittle:
Ping @rcomer :wink:

rcomer:
Great news!  Thanks both for your work on this :grin:

PR #141, opened 2019-03-05
corrected python-eccodes requirement in install_requires

corinnebosley:
Replaced 'eccodes' with 'python-eccodes' in setup.py install_requires section.  Necessary for installation success.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,7 @@ from __future__ import print_function
 
 import os
 import os.path
+import sys
 from setuptools import setup
 import textwrap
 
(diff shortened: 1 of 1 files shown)


corinnebosley:
@bjlittle do you want me to create an issue outlining the details of this PR and how to correct it for the next release?  Associated with a milestone, maybe...?

bjlittle:
@corinnebosley Thanks. The `pip` dependencies are pretty restrictive at the moment for `eccodes`, and there is not much we can do about that at the moment, but temporarily side-step the issue here

corinnebosley:
Thanks @bjlittle!  Issue created, can't attach milestone for next release because it doesn't exist yet and I'm not sure how to make it or even if I can.  It's there anyway.

PR #149, opened 2019-07-09
Test most recent eccodes release

lbdreyer:
Eccodes v2.13.0 was just released, so this is to test iris-grib with that release.
Furthermore, eccodes-python has been updated such that things like `grib_get_double_array` returns arrays as we would expect, rather than a list, and as such I have pulled out those related workarounds.

FYI @bjlittle

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ install:
     
   # If running in Python 3, install eccodes-python and check the installation.
   - if [[ "${python}" != "2.7" ]]; then
-        pip install eccodes-python;
+        pip install eccodes-python>=0.9.1;
         python -m eccodes selfcheck;
     fi
 
--- a/iris_grib/__init__.py
+++ b/iris_grib/__init__.py
@@ -44,7 +44,7 @@ from iris.exceptions import TranslationError, NotYetImplementedError
 from . import grib_phenom_translation as gptx
 from . import _save_rules
 from ._load_convert import convert as load_convert
-from .message import GribMessage, _grib_get_array_wrapper
+from .message import GribMessage
 
 
 __version__ = '0.14.1'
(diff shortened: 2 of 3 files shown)


bjlittle:
@lbdreyer Awesome, thanks.

So this pretty much undoes the temporary workarounds that we banked in #138, right?

Also, I'm assuming that we still need the `int` casting? See [here](https://github.com/SciTools/iris-grib/pull/138/files#diff-cb7c8f2df28be9778885dcaff9827a70R980) in #138

lbdreyer:
> So this pretty much undoes the temporary workarounds that we banked in #138, right?
>
>Also, I'm assuming that we still need the int casting? See here in #138

Yes this undoes the temporary changes in #138 that were there because eccodes-python was returning a list when it should have been returning an array. I have also pushed up an extra commit that now pins the version of eccodes-python that we install as the fix was added in 0.9.1.

The only changes from #138 that I haven't undone are, as you say the int casting which is due to a change in eccodes itself and is present in Python 2 as well as Python 3. The other change in #138 that I haven't undone is the change of
`gribapi.GribInternalError` -> `gribapi.errors.GribInternalError` 
(see [here](https://github.com/SciTools/iris-grib/pull/138/files#diff-b4b298fb02976da69d34fbfe378ab0feR229))
as that was also required due a change in eccodes.

bjlittle:
@lbdreyer LGTM

You haven't got any more changes, right? Otherwise, I'm happy to merge...

lbdreyer:
@bjlittle No more changes :)

Thanks for looking at this!

bjlittle:
@lbdreyer Did we want to aim for a `0.14.2` release, or I'm thinking a `0.15.0` release and include other PRs now that we have the opportunity?

lbdreyer:
I'd prefer a `0.15.0` release. 
In any case, @pp-mo has found a few issues when testing Iris with the latest eccode. It turns out there are few tests in Iris that haven't made their way over to iris-grib yet. So when @pp-mo ran the Iris tests with the latest version of eccodes, he found a few things in iris-grib that need to be fixed.
I'd like to get these fixed soonish, and hopefully put out with the 0.15 release

bjlittle:
Cool. I'm already on his `iris` changes... `0.15.0` makes sense to me.

Issue #153, opened 2019-08-02
Provide format-specific phenomenon identity for load and save


pp-mo:
Summary : a solution like the 'STASH' attribute, to carry grib-specific parameter identity.

At present, if you want to save Iris data for which no known parameter code translation exists, it just gets a 0, so you can't save multiple such things to one file.

Likewise on load, an unknown parameter code just results in an 'unknown' or anonymous cube 
 -- and maybe we can even wind up merging things that should be kept separate ??

I'm imagining a specially recognised cube attribute, along the lines of 'STASH' for PP+FF encoding.
It needs to encode : (edition, discipline parameterCategory, parameterNumber).  We can probably handle GRIB1 in that form somehow, not clear on detail yet.

pp-mo:
Fixed by #156

PR #156, opened 2019-09-06
Label unknown phenomena with a 'GRIB_CODING' attribute

pp-mo:
Aims to address #153

### DIFF
--- a/iris_grib/_load_convert.py
+++ b/iris_grib/_load_convert.py
@@ -26,13 +26,13 @@ import iris.coord_systems as icoord_systems
 from iris.coords import AuxCoord, DimCoord, CellMethod
 from iris.exceptions import TranslationError
 from . import grib_phenom_translation as itranslation
+from .grib_phenom_translation import GribCode
 from iris.fileformats.rules import ConversionMetadata, Factory, Reference, \
     ReferenceTarget
 from iris.util import _is_circular
 
 from ._iris_mercator_support import confirm_extended_mercator_supported
 from ._grib1_load_rules import grib1_convert
-from .message import GribMessage
 
 
 # Restrict the names imported from this namespace.
@@ -1357,7 +1357,15 @@ def translate_phenomenon(metadata, discipline, parameterCategory,
     cf = itranslation.grib2_phenom_to_cf_info(param_discipline=discipline,
                                               param_category=parameterCategory,
                                               param_number=parameterNumber)
-    if cf is not None:
+    if cf is None:
+        # Add a standard attribute recording the grib phenomenon identity.
+        # Note: deferred import to avoid circular import problem
+        metadata['attributes']['GRIB_CODING'] = GribCode(
+            edition_or_string=2,
+            discipline=discipline,
+            category=parameterCategory,
+            number=parameterNumber)
+    else:
         if probability is None:
             metadata['standard_name'] = cf.standard_name
             metadata['long_name'] = cf.long_name
--- a/iris_grib/_load_convert.py
+++ b/iris_grib/_load_convert.py
@@ -26,7 +26,7 @@ import iris.coord_systems as icoord_systems
 from iris.coords import AuxCoord, DimCoord, CellMethod
 from iris.exceptions import TranslationError
 from . import grib_phenom_translation as itranslation
-from .grib_phenom_translation import GRIBcode
+from .grib_phenom_translation import GRIBCode
 from iris.fileformats.rules import ConversionMetadata, Factory, Reference, \
     ReferenceTarget
 from iris.util import _is_circular
@@ -1379,7 +1379,7 @@ def translate_phenomenon(metadata, discipline, parameterCategory,
             metadata['units'] = Unit(1)
 
     # Add a standard attribute recording the grib phenomenon identity.
-    metadata['attributes']['GRIB_PARAM'] = GRIBcode(
+    metadata['attributes']['GRIB_PARAM'] = GRIBCode(
         edition_or_string=2,
         discipline=discipline,
         category=parameterCategory,
--- a/iris_grib/_load_convert.py
+++ b/iris_grib/_load_convert.py
@@ -1357,14 +1357,7 @@ def translate_phenomenon(metadata, discipline, parameterCategory,
     cf = itranslation.grib2_phenom_to_cf_info(param_discipline=discipline,
                                               param_category=parameterCategory,
                                               param_number=parameterNumber)
-    if cf is None:
-        # Add a standard attribute recording the grib phenomenon identity.
-        metadata['attributes']['GRIB_CODING'] = GribCode(
-            edition_or_string=2,
-            discipline=discipline,
-            category=parameterCategory,
-            number=parameterNumber)
-    else:
+    if cf is not None:
         if probability is None:
             metadata['standard_name'] = cf.standard_name
             metadata['long_name'] = cf.long_name
@@ -1384,6 +1377,13 @@ def translate_phenomenon(metadata, discipline, parameterCategory,
             metadata['standard_name'] = None
             metadata['long_name'] = long_name
             metadata['units'] = Unit(1)
+
+    # Add a standard attribute recording the grib phenomenon identity.
+    metadata['attributes']['GRIB_CODING'] = GribCode(
+        edition_or_string=2,
+        discipline=discipline,
+        category=parameterCategory,
+        number=parameterNumber)
 
     # Identify hybrid height and pressure reference fields.
     # Look for fields at surface level first.
--- a/iris_grib/_load_convert.py
+++ b/iris_grib/_load_convert.py
@@ -26,7 +26,7 @@ import iris.coord_systems as icoord_systems
 from iris.coords import AuxCoord, DimCoord, CellMethod
 from iris.exceptions import TranslationError
 from . import grib_phenom_translation as itranslation
-from .grib_phenom_translation import GribCode
+from .grib_phenom_translation import GRIBcode
 from iris.fileformats.rules import ConversionMetadata, Factory, Reference, \
     ReferenceTarget
 from iris.util import _is_circular
@@ -1379,7 +1379,7 @@ def translate_phenomenon(metadata, discipline, parameterCategory,
             metadata['units'] = Unit(1)
 
     # Add a standard attribute recording the grib phenomenon identity.
-    metadata['attributes']['GRIB_CODING'] = GribCode(
+    metadata['attributes']['GRIB_PARAM'] = GRIBcode(
         edition_or_string=2,
         discipline=discipline,
         category=parameterCategory,
--- a/iris_grib/_load_convert.py
+++ b/iris_grib/_load_convert.py
@@ -1359,7 +1359,6 @@ def translate_phenomenon(metadata, discipline, parameterCategory,
                                               param_number=parameterNumber)
     if cf is None:
         # Add a standard attribute recording the grib phenomenon identity.
-        # Note: deferred import to avoid circular import problem
         metadata['attributes']['GRIB_CODING'] = GribCode(
             edition_or_string=2,
             discipline=discipline,
--- a/iris_grib/_save_rules.py
+++ b/iris_grib/_save_rules.py
@@ -712,7 +712,6 @@ def set_discipline_and_parameter(cube, grib):
         if attr:
             try:
                 # Convert to standard tuple-derived form.
-                # Late import to avoid circularity
                 gc = GribCode(attr)
                 if gc.edition == 2:
                     discipline = gc.discipline
--- a/iris_grib/_save_rules.py
+++ b/iris_grib/_save_rules.py
@@ -30,7 +30,7 @@ from ._iris_mercator_support import confirm_extended_mercator_supported
 from . import grib_phenom_translation as gptx
 from ._load_convert import (_STATISTIC_TYPE_NAMES, _TIME_RANGE_UNITS,
                             _SPATIAL_PROCESSING_TYPES)
-from .grib_phenom_translation import GribCode
+from .grib_phenom_translation import GRIBcode
 from iris.util import is_regular, regular_step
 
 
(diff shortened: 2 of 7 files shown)


iris_grib/_load_convert.py
```diff
@@ -1341,7 +1341,15 @@ def translate_phenomenon(metadata, discipline, parameterCategory,
     cf = itranslation.grib2_phenom_to_cf_info(param_discipline=discipline,
                                               param_category=parameterCategory,
                                               param_number=parameterNumber)
-    if cf is not None:
+    if cf is None:
+        # Add a standard attribute recording the grib phenomenon identity.
+        # Note: deferred import to avoid circular import problem
```
pp-mo: Not so! 
obsolete comment

iris_grib/_save_rules.py
```diff
@@ -704,22 +705,46 @@ def grid_definition_section(cube, grib):
...
+        number = grib2_info.number
+        identity_found = True
+
+    if not identity_found:
+        # See if we can find and interpret a 'GRIB_CODING' attribute.
+        attr = cube.attributes.get('GRIB_CODING', None)
+        if attr:
+            try:
+                # Convert to standard tuple-derived form.
+                # Late import to avoid circularity
```
pp-mo: Again, not so -- remove obsolete comment

iris_grib/grib_phenom_translation.py
```diff
@@ -331,3 +332,59 @@ def cf_phenom_to_grib2_info(standard_name, long_name=None):
...
+    """
+    __slots__ = ()
+
+    def __new__(cls, edition_or_string,
+                discipline=None, category=None, number=None):
+        if (all(arg is None
+                for arg in (discipline, category, number))):
+            # Only 1 arg : can  be existing object, or a string.
+            # Convert to a string : also covers case of a GribCode to copy.
+            edition_or_string = str(edition_or_string)
```
pp-mo: Note: a tuple of numbers would also work here.
Worth adding to comment

iris_grib/grib_phenom_translation.py
```diff
@@ -331,3 +332,59 @@ def cf_phenom_to_grib2_info(standard_name, long_name=None):
...
+
+    """
+    __slots__ = ()
+
+    def __new__(cls, edition_or_string,
+                discipline=None, category=None, number=None):
+        if (all(arg is None
+                for arg in (discipline, category, number))):
+            # Only 1 arg : can  be existing object, or a string.
+            # Convert to a string : also covers case of a GribCode to copy.
```
pp-mo: Note: a tuple of numbers will also work here via the str() route.
Worth adding to comment

iris_grib/grib_phenom_translation.py
```diff
@@ -331,3 +332,59 @@ def cf_phenom_to_grib2_info(standard_name, long_name=None):
...
+                for arg in (edition, discipline, category, number)]
+        return super(GribCode, cls).__new__(
+            cls, edition, discipline, category, number)
+
+    RE_PARSE_FOURNUMS = re.compile(4 * r'[^\d]*(\d*)')
+
+    @classmethod
+    def _fournums_from_gribcode_string(cls, edcn_string):
+        nums_match = cls.RE_PARSE_FOURNUMS.match(edcn_string).groups()
+        parsed_ok = nums_match is not None
```
pp-mo: A "set-once, default is False" strategy for 'parsed_ok' would be clearer.

pp-mo:
@ehogan I think you might be interested in this ?

pp-mo:
Good to go now AFAIK !

@bjlittle @lbdreyer a distraction from Iris, but can you summon any interest in this ?

pp-mo:
Some practical notes:

This is supposed to work just like the "STASH" attribute that Iris uses for loading + saving UM files into Iris (i.e. fieldsfiles or PP).
However.. when loading UM files, you *always* get a "STASH" attribute,
but when loading grib data, you will now *only* get a "GRIB_CODING" attribute when the phenomenon is unrecognised.

This is intended to avoid confusing existing loading code.  
However, it will lead to results changing when we add new phenomenon codings.  So that may not be ideal.

A safer way, more consistent with general Iris usage, would be ...
 * use a FUTURE type flag to enable GRIB_CODING attributes for all loads
  * and deprecate current usage
  * tidy up at a major release 

But that's overkill for iris-grib,
isn't it ... ???

pp-mo:
From discussion with @bjlittle ...

> A safer way, ... use a FUTURE type flag 
> that's overkill for iris-grib,  isn't it ... ???

Two points here :
  * (1) the above fuss is strictly ***not*** required, because we don't have a version 1.0 yet
  * (2) the existing proposal is still a bit flaky + not that safe for some reasons ...
    * adding a cf-translation for a code means it will stop loading with a grib-coding attribute 
    * unlike Iris-STASH-attribute, the control attribute does not override on save, but is only honoured when a cf-translation is absent.

Given (1), we're probably better off implementing a mechanism identical to the iris-STASH-attribute, 
as this is easier to explain + understand.  That means ...
  * on load, **all** data will have the new attribute
  * on save, the presence of a stash-coding attribute will set the encoding + ignore a cf-translation

I'll just run off + implement that ...

pp-mo:
Update : 
Just rebased
Still needs some test changes/fixes to accommodate the recent policy change...

iris_grib/tests/unit/save_rules/test_set_discipline_and_parameter.py
```diff
@@ -0,0 +1,88 @@
...
+# Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# iris-grib is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with iris-grib.  If not, see <http://www.gnu.org/licenses/>.
```
lbdreyer: You should also update this license header

pp-mo: Oops.
Let's see if the new header check finds it !

pp-mo: yup! :1st_place_medal:

pp-mo:
Thanks @lbdreyer 
Adopted revised policy + fixed the PR.
Now good to go I think, please review ...

iris_grib/tests/unit/save_rules/test_set_discipline_and_parameter.py
```diff
@@ -0,0 +1,83 @@
...
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
+"""Unit tests for `iris_grib.grib_save_rules.set_discipline_and_parameter`."""
+from __future__ import (absolute_import, division, print_function)
+from six.moves import (filter, input, map, range, zip)  # noqa
+
+# Import iris_grib.tests first so that some things can be initialised before
+# importing anything else.
+import iris_grib.tests as tests
+import mock
```
lbdreyer:
```suggestion
from unittest import mock
```

iris_grib/_save_rules.py
```diff
@@ -693,22 +694,45 @@ def grid_definition_section(cube, grib):
...
-        gribapi.grib_set(grib, "parameterNumber", grib2_info.number)
-    else:
-        gribapi.grib_set(grib, "discipline", 255)
-        gribapi.grib_set(grib, "parameterCategory", 255)
-        gribapi.grib_set(grib, "parameterNumber", 255)
+    # Default values for parameter identity keys = effectively "MISSING".
+    discipline, category, number = 255, 255, 255
+    identity_found = False
+
+    # First, see if we can find and interpret a 'GRIB_CODING' attribute.
```
lbdreyer: I'm not convinced its a good idea to interpret the "GRIB_CODING" attribute first.
I think the standard_name/long_name should take precedence when determining the correct discipline/category/number as that it is the primary means we have of identifying a cube. This reliance on GRIB_CODING means in future we have to take extra care with attributes and I don't want that requirement. 
For example someone take a copy of a cube, update the data array, and only update the name, then save it out as a cube

iris_grib/grib_phenom_translation.py
```diff
@@ -320,3 +321,70 @@ def cf_phenom_to_grib2_info(standard_name, long_name=None):
...
+
+        if not parsed_ok:
+            msg = ('Invalid argument for GribCode creation, '
+                   '"GribCode({!r})" : '
+                   'requires 4 numbers, separated by non-numerals.')
+            raise ValueError(msg.format(edcn_string))
+
+        return nums
+
+    PRINT_FORMAT = 'GRIB{:1d}:d{:d}.c{:d}.n{:d}'
```
lbdreyer: Rather than
` 'GRIB2:d1.c1.n1`
How about
`GRIB2:d001c001n001`
or
 `GRIB2:d1c1n1`

iris_grib/_save_rules.py
lbdreyer: From an offline discussion with @pp-mo, it turns out that this is how the STASH attribute is handled :warning: :warning: :warning: 

In which case, my preference is for matching that behaviour so I'm happy with this as it is.

iris_grib/_load_convert.py
```diff
@@ -1380,6 +1380,13 @@ def translate_phenomenon(metadata, discipline, parameterCategory,
             metadata['long_name'] = long_name
             metadata['units'] = Unit(1)
 
+    # Add a standard attribute recording the grib phenomenon identity.
+    metadata['attributes']['GRIB_CODING'] = GribCode(
```
lbdreyer: I think `GRIB_CODING` lacks clarity on what it is referring to, and that it needs some reference to "parameter"
I would prefer something like
`GRIB_PARAM`
or 
`GRIB_PARAM_CODE`

lbdreyer:
After more thinking and discussing, I propose the following format for the cube attribute:
```
attributes  = {
'GRIB_PARAM': 'GRIB2:d001c001n001'}
```
Where `GRIB_PARAM` is the key, and the value follows the format `GRIB?:d???c???n???`

iris_grib/grib_phenom_translation.py
pp-mo: Will fix to follow https://github.com/SciTools/iris-grib/pull/156#issuecomment-560362457 suggestions

iris_grib/grib_phenom_translation.py
```diff
@@ -19,9 +19,10 @@
 
 from __future__ import (absolute_import, division, print_function)
 from six.moves import (filter, input, map, range, zip)  # noqa
-import six
 
-import collections
+from collections import namedtuple
+import re
+import six
```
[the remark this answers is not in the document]
pp-mo: Ok

iris_grib/tests/unit/grib_phenom_translation/test_grib_phenom_translation.py
```diff
@@ -17,9 +17,12 @@
 # importing anything else.
 import iris_grib.tests as tests
 
+import six
```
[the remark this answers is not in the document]
pp-mo: Ok

iris_grib/tests/unit/save_rules/test_set_discipline_and_parameter.py
```diff
@@ -0,0 +1,83 @@
+# Copyright iris-grib contributors
+#
+# This file is part of iris-grib and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
+"""Unit tests for `iris_grib.grib_save_rules.set_discipline_and_parameter`."""
+from __future__ import (absolute_import, division, print_function)
+from six.moves import (filter, input, map, range, zip)  # noqa
```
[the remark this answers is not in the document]
pp-mo: Ok. Nicer !

pp-mo:
Rebased to fix.
Other requested changes coming ...

pp-mo:
Thanks @lbdreyer.
Should be good now ?
Note that I thought of + added an extra test.

iris_grib/grib_phenom_translation.py
```diff
@@ -320,3 +321,70 @@ def cf_phenom_to_grib2_info(standard_name, long_name=None):
     if standard_name is not None:
         long_name = None
     return _CF_GRIB2_TABLE[(standard_name, long_name)]
+
+
+class GribCode(namedtuple('GribCode',
```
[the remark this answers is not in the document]
pp-mo: I tried  'GRIBcode'.  Is that ok ?

iris_grib/tests/unit/grib_phenom_translation/test_grib_phenom_translation.py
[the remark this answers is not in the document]
lbdreyer: This still needs removing?

iris_grib/grib_phenom_translation.py
```diff
@@ -316,3 +316,70 @@ def cf_phenom_to_grib2_info(standard_name, long_name=None):
     if standard_name is not None:
         long_name = None
     return _CF_GRIB2_TABLE[(standard_name, long_name)]
+
+
+class GRIBcode(namedtuple('GRIBcode',
```
lbdreyer: I'm afraid I don't like `GRIBcode` I read that as `GRI Bcode`

Maybe it is best to go back to `GribCode` if you don't like GRIBCode
Wikipedia's article on camel links to an paper that points out a run of uppercase letters can cause problems which I think we're having with GRIBCode. So I don't think we have to strictly treat acronyms as all capital letters.

iris_grib/tests/unit/grib_phenom_translation/test_grib_phenom_translation.py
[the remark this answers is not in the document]
pp-mo: Sorry I got confused + thought I'd done this.  Will fix.

iris_grib/grib_phenom_translation.py
pp-mo: > if you don't like GRIBCode

I read as "GRIBC ode".  It seems to be the opposite problem to your "GRI Bcode" !

pp-mo:
Hi @lbdreyer.
Changed as requested.
Can we go with this ?

lbdreyer:
> Hi @lbdreyer.
> Changed as requested.
> Can we go with this ?

Hi @pp-mo 

Yep this is _almost_ good to go. Just need you to get rid of [the six import](https://github.com/SciTools/iris-grib/pull/156/files#r352154757)

pp-mo:
here it is, @lbdreyer !
Sorry for another mistake + delay.

lbdreyer:
:tada: six is now removed :tada:


I'll merge this in ....

PR #165, opened 2019-11-26
Revert pv level indexing change

### LINKED ISSUE #150: Fix level indexing in hybrid coords
Revert #106 !
I think this change was a misunderstanding : the test data does not show clearly that these were the wrong numbers -- but other (real) data since encountered does.

( N.B. the actual Grib spec is really vague on this, hence the interpretation depended on what I found in actual files ! )

lbdreyer:
Closes #150 

I have also convinced myself that the indexing before was correct, based on looking at some data produced from ECMWF, rather than our own sample data

@pp-mo Are you able to review?

### DIFF
--- a/iris_grib/_load_convert.py
+++ b/iris_grib/_load_convert.py
@@ -1511,6 +1511,9 @@ def hybrid_factories(section, metadata):
                                 Reference('ref_surface_pressure')]
 
             # Create the level height/pressure scalar coordinate.
+            # scaledValue represents the level number, which is used to select
+            # the sigma and delta values as follows:
+            # sigma, delta = PV[i], PV[NV/2+i] : where i=1..level_number
             pv = section['pv']
             offset = scaledValue
             coord = DimCoord(pv[offset], long_name=level_value_name,
--- a/iris_grib/_load_convert.py
+++ b/iris_grib/_load_convert.py
@@ -1510,14 +1510,14 @@ def hybrid_factories(section, metadata):
                                 {'long_name': 'sigma'},
                                 Reference('ref_surface_pressure')]
 
-            # Create the level pressure scalar coordinate.
+            # Create the level height/pressure scalar coordinate.
             pv = section['pv']
-            offset = scaledValue - 1
+            offset = scaledValue
             coord = DimCoord(pv[offset], long_name=level_value_name,
                              units=level_value_units)
             metadata['aux_coords_and_dims'].append((coord, None))
             # Create the sigma scalar coordinate.
-            offset = NV // 2 + scaledValue - 1
+            offset = NV // 2 + scaledValue
             coord = AuxCoord(pv[offset], long_name='sigma')
             metadata['aux_coords_and_dims'].append((coord, None))
             # Create the associated factory reference.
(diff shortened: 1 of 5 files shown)


iris_grib/_load_convert.py
```diff
@@ -1492,14 +1492,14 @@ def hybrid_factories(section, metadata):
...
+            # Create the level height/pressure scalar coordinate.
             pv = section['pv']
-            offset = scaledValue - 1
+            offset = scaledValue
             coord = DimCoord(pv[offset], long_name=level_value_name,
                              units=level_value_units)
             metadata['aux_coords_and_dims'].append((coord, None))
             # Create the sigma scalar coordinate.
-            offset = NV // 2 + scaledValue - 1
+            offset = NV // 2 + scaledValue
```
lbdreyer: Note! prior to #106 this would have been `scaledValue + NV // 2` but I don't think that is correct

iris_grib/_save_rules.py
```diff
@@ -953,9 +953,9 @@ def set_fixed_surfaces(cube, grib, full3d_cube=None):
                    'maximum value = {}.')
             raise ValueError(msg.format(max_valid_nlevels, n_levels))
         # In sample data we have seen, there seems to be an extra missing data
-        # value after each set of n-levels coefficients.
-        # For now, we retain that encoding. This should not cause problems as
-        # values are indexed according to model-level-number,
+        # value *before* each set of n-levels coefficients.
+        # Note: values are indexed according to model_level_number, using
+        # zero-based numbering,
```
pp-mo: I think the 'using zero-based numbering' bit is still confusing!
You might add to the following ..
`# I.E. sigma, delta = PV[i], PV[NV/2+i]  : where i=1..n_levels`

iris_grib/_load_convert.py
```diff
@@ -1510,14 +1510,14 @@ def hybrid_factories(section, metadata):
                                 {'long_name': 'sigma'},
                                 Reference('ref_surface_pressure')]
 
-            # Create the level pressure scalar coordinate.
+            # Create the level height/pressure scalar coordinate.
             pv = section['pv']
-            offset = scaledValue - 1
+            offset = scaledValue
```
pp-mo: Maybe worth an extra comment here also, about 'scaledValue' being 1-based, or "1..n_levels".
And/or maybe rename the variable 'level_number' ??

pp-mo:
Thanks @lbdreyer 
Function is good.  Just a couple of ideas for making code a bit clearer.
Then will merge !

iris_grib/_load_convert.py
lbdreyer: I don't agree with renaming the variable `level_number` as offset is meant to be the offset within the the pv array, see line 1520 where the offset is different

lbdreyer:
@pp-mo I have addressed your comments in the most recent commit. 
I had a little difficulty coming up with good commits so let me know if you can think of a better way to explain it.

iris_grib/_load_convert.py
pp-mo: Sorry, I meant that you could rename ***`scaledValue`*** as `level_number`.  I.E. what you get as the principal coordinate value is in this case a level number, i.e. a 1..N value.

I will check out your other changes..

pp-mo:
> let me know if you can think of a better way to explain it.

No, this does it for me, I think that _is_ clear.

pp-mo:
Thanks @lbdreyer !

PR #161, opened 2019-10-23
Depth coordinate

BHFock:
Add depth variable to allow GRIB conversion of surface fields with depth coordinate, e.g. of UM fieldsfiles for soil moisture (lbuser4=8223) or soil temperature (lbuser4=8225).

PR #151, opened 2019-07-12
update travis

bjlittle:
This PR tidies the `.travis.yml` and extends it to provide Travis CI test coverage of `python` version `3.7`.

It also includes the switch to using minimal Travis CI images.

PR #148, opened 2019-06-25
Fix to read grib files obeying 'resolutionAndComponentFlags'

tpowellmeto:
This fix allows iris to obey the 'resolutionAndComponentFlags' value within GRIB files. This value specifies whether iris should be looking for i/j increments within the given GRIB file, and if they aren't specified (value of 0 is set) provides a method of calculating the increments via other means.

PR #146, opened 2019-06-04
Granularity for fixed surface arguments

trexfeathers:
product_definition_section will no longer always try to populate the fixed surface arguments for translate_phenomenon - not available in some templates, in which case populate with None

Ran iris-grib unit tests afterwards - 364 passed, 21 were ignored as they needed external data (total 385)

PR #166, opened 2019-11-26
Drop Python 2 Testing

lbdreyer:
I think we should consider dropping Python 2 support. 

Due to the confusing nature of Python 2/3 support from `eccodes`/`eccodes-python`/`python-eccodes`, having Python 2 and Python 3 support means that the recipe/requirements will be horribly complicated.
We could simplify this by dropping Python 2. 
And anyway, Python 2 EoL is around the corner!! 

Dropping Python 2 support would mean we would have to make the next release a major release (iris grib 1.0) and remove all the Python 2 stuff (six etc)

Thoughts @bjlittle @pp-mo ?

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -2,4 +2,4 @@ channels:
     - conda-forge
 dependencies:
     - iris=2.*
-    - python-eccodes
+    - python-eccodes=0.9.3
--- a/environment.yml
+++ b/environment.yml
@@ -2,4 +2,4 @@ channels:
     - conda-forge
 dependencies:
     - iris=2.*
-    - python-eccodes=0.9.3
+    - python-eccodes>=0.9.1,<2
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ dist: xenial
 
 env:
   matrix:
-    - PYTHON_VERSION=2.7
     - PYTHON_VERSION=3.6
     - PYTHON_VERSION=3.7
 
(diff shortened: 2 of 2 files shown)


environment.yml
```diff
@@ -2,4 +2,4 @@ channels:
     - conda-forge
 dependencies:
     - iris=2.*
-    - python-eccodes
```
lbdreyer: Have to pin this as otherwise it will pick up `python-eccodes 2.12.3` which is Python 2. 
`python-eccodes 0.9.3`, confusingly, is Python 3 only

lbdreyer:
@pp-mo Below is a list of what has gone in since the 0.14.1 release:

* Updated metadata translations #143 
* Fixes to travis including the addition of Python 3.7 testing #151 
* Bug fix: Granularity for fixed surface arguments #146 
* Supporting saving out depth coordinates #161 
* Support loading coordinates from files without an x/y increment #148 

And any thing else we manage to fit into this release.
I think we could justify not releasing the above in Python 2.

environment.yml
pp-mo: :tired_face:  :nauseated_face: :sneezing_face:
!

pp-mo: I think we should try to leave this a bit flexible.
Can we have ">0.9 <2" ??

pp-mo: @lbdreyer actually I think this needs to be ">=0.9.1",
to accommodate the removal of certain workarounds, which #169 introduces.
(as on 14.x branch, this stuff was added in #138, removed in #149).

pp-mo:
:+1: LGTM, getting python-eccodes v0.9.3 which I guess is a win !

PR #167, opened 2019-11-26
Update license headers

lbdreyer:
If we are going to drop Python 2 support, that will mean touching almost all files, so we may as well update the license headers first as done in Iris (see https://github.com/SciTools/iris/pull/3489/)


lbdreyer:
Thanks @pp-mo !!

PR #169, opened 2019-11-27
0.14.x mergeback onto master

lbdreyer:
A mergeback has not been done since the 0.14.1 release and so the changes included in #149 needs to be merged back onto master

### DIFF
--- a/iris_grib/__init__.py
+++ b/iris_grib/__init__.py
@@ -29,7 +29,7 @@ from iris.exceptions import TranslationError, NotYetImplementedError
 from . import grib_phenom_translation as gptx
 from . import _save_rules
 from ._load_convert import convert as load_convert
-from .message import GribMessage, _grib_get_array_wrapper
+from .message import GribMessage
 
 
 __version__ = '0.15.0dev0'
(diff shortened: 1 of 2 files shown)


lbdreyer:
@pp-mo are you available to have a look at this?

lbdreyer:
Closes #168

pp-mo:
Thanks @lbdreyer 
Looking at this, it has only 2 commits.
  * the [first commit](https://github.com/SciTools/iris-grib/pull/169/commits/9446e083968b4198b7eebc08b95daa068b661b97) removes workarounds :  all looks fine
  * the [second commit](https://github.com/SciTools/iris-grib/pull/169/commits/17664b55a48a1e3953db91273e9bd1d7e4bee957) pins eccodes version, **which I think will simply conflict with #166**

So I would recommend waiting until #166, then we just use the first commit.
**Unless** you think we need a complete merge-back to enable future developments on 0.14.x ???

lbdreyer:
> So I would recommend waiting until #166, then we just use the first commit.
Unless you think we need a complete merge-back to enable future developments on 0.14.x ???

Happy with this :)
I don't believe a complete merge-back is necessary for 0.14.x dev.

pp-mo:
Hi @lbdreyer 
Since #166 this should be ok, but it needs 
  1. a rebase and 
  2. include only the first of 2 commits, as discussed.

lbdreyer:
Thanks @pp-mo! I have now updated this such that it only includes the first commit

lbdreyer:
I have now resolved the conflicts. Just need the tests to pass then this can go in!

PR #170, opened 2019-11-27
Drop Python 2 support

lbdreyer:
This removes (hopefully) all Python 2 support from iris_grib.

I have checked for :
* six
* `__future__`
* class <name_of_class>(object):
* super arguments
* mock rather than unittest.mock
* `assertRaisesRegexp` > `assertRaisesRegex`

Note ! I have not touched [_grib_cf_map.py](https://github.com/SciTools/iris-grib/blob/master/iris_grib/_grib_cf_map.py) as that file is auto-generated and so should not be manually modified

### DIFF
--- a/iris_grib/_load_convert.py
+++ b/iris_grib/_load_convert.py
@@ -10,7 +10,8 @@ cube metadata.
 """
 
 from argparse import Namespace
-from collections import namedtuple, Iterable, OrderedDict
+from collections import namedtuple, OrderedDict
+from collections.abc import Iterable
 from datetime import datetime, timedelta
 import math
 import warnings
--- a/iris_grib/_load_convert.py
+++ b/iris_grib/_load_convert.py
@@ -9,9 +9,6 @@ cube metadata.
 
 """
 
-from __future__ import (absolute_import, division, print_function)
-from six.moves import (filter, input, map, range, zip)  # noqa
-
 from argparse import Namespace
 from collections import namedtuple, Iterable, OrderedDict
 from datetime import datetime, timedelta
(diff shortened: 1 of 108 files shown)


iris_grib/_load_convert.py
```diff
@@ -9,9 +9,6 @@
 
 """
 
-from __future__ import (absolute_import, division, print_function)
-from six.moves import (filter, input, map, range, zip)  # noqa
-
 from argparse import Namespace
 from collections import namedtuple, Iterable, OrderedDict
```
pp-mo: I think some of these have "moved" :  Iterable should now be imported from `collections.abc`.
(`namedtuple` and `OrderedDict` are ok)

I'm not sure why this doesn't give a deprecation warning in the Travis log : maybe because it is not "strictly" a Python3 migration issue as it was supposedly "better" to do it this way before ??

According to [this note](https://docs.python.org/3/library/collections.html), this affects _all_ the ABCs like Callable, Iterable etc, and is "Deprecated since version 3.3, will be removed in version 3.9".

I will check for these + raise separate comments elsewhere if I find any ...

pp-mo:
Thanks @lbdreyer 
I checked for all the things done in https://github.com/SciTools/iris/pull/3513.
Other than the tiny point just raised, I think this is all :+1: :1st_place_medal: :heavy_check_mark: !

pp-mo:
Thanks @lbdreyer 
I checked for all the things done in https://github.com/SciTools/iris/pull/3513.
Other than the tiny point just raised, I think this is all :+1: :1st_place_medal: :heavy_check_mark: !

PR #160, opened 2019-10-23
Add cf to grib2 mapping for cloud ice and cloud water

BHFock:
Add CF to GRIB2 mapping to improve GRIB header when converting UM FieldsFile to GRIB2. The added mappings are:

- mass_fraction_of_cloud_liquid_water_in_air to  "specific cloud liquid water content", http://codes.wmo.int/grib2/codeflag/4.2/_0-1-83 

- Map mass_fraction_of_cloud_ice_in_air to "specific cloud ice water content", 
http://codes.wmo.int/grib2/codeflag/4.2/_0-1-84|

This change improves the GRIB headers when converting UM field files with the variables lbuser4=254 (QCL AFTER TIMESTEP) and lbuser4=12 (QCF AFTER TIMESTEP) to GRIB2.

PR #172, opened 2019-11-28
https://github.com/metarelate/metOcean/issues/95

marqh:
update following:
https://github.com/metarelate/metOcean/issues/95

closes #160

### DIFF
--- a/iris_grib/_grib_cf_map.py
+++ b/iris_grib/_grib_cf_map.py
@@ -16,11 +16,11 @@
 # along with iris-grib.  If not, see <http://www.gnu.org/licenses/>.
 #
 # DO NOT EDIT: AUTO-GENERATED
-# Created on 22 May 2019 13:05 from 
+# Created on 28 November 2019 16:11 from 
 # http://www.metarelate.net/metOcean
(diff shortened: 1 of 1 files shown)


iris_grib/_grib_cf_map.py
```diff
@@ -34,6 +34,7 @@
```
lbdreyer: We have recently dropped Python 2 support and so have been removing any use of six/__future__

Would you be able to also remove lines 30 and 31 as they are no longer needed?

iris_grib/_grib_cf_map.py
```diff
@@ -16,11 +16,11 @@
 # along with iris-grib.  If not, see <http://www.gnu.org/licenses/>.
```
lbdreyer: Could you also edit the license header to be the same as the other files:
https://github.com/SciTools/iris-grib/blob/master/iris_grib/__init__.py#L1-L5

We have dropped the old complex license header.

This change will also apply to iris

lbdreyer:
This is great! Thanks @marqh !

Note: I have put a PR to update the gen_translations.py  tool in Iris to reflect the simpler license headers (see https://github.com/SciTools/iris/pull/3567)

PR #184, opened 2020-01-08
Latest Iris Compatibility

trexfeathers:
This PR makes some minor changes that allow the latest Iris-GRIB to be installed alongside the latest Iris, including ensuring that the Iris-GRIB tests can run.

_Note that Iris-GRIB includes a pep8 test, but pep8 is not included in [environment.yml](https://github.com/SciTools/iris-grib/blob/master/environment.yml) since it is not a core requirement. In the past, pep8 would be included with a full Iris install, but that is no longer the case since Iris was switched to black syntax, so running the tests from the command line may error on the pep8 test depending on the user's install. pep8 is however installed during the Travis tests._

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,5 +1,5 @@
 channels:
     - conda-forge
 dependencies:
-    - iris=2.*
+    - iris>=2
     - python-eccodes>=0.9.1,<2
--- a/environment.yml
+++ b/environment.yml
@@ -3,3 +3,4 @@ channels:
 dependencies:
     - iris>=2
     - python-eccodes>=0.9.1,<2
+    - pep8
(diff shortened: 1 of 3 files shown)


trexfeathers:
**NOTE:** since Travis relies on installing Iris from conda, this PR will not pass testing until the move of `equalise_attributes` to `iris.util` is within an Iris version that is available on conda. Merging will need to wait until then.

Depending on release timings, it may be necessary to extract the fix for `test_grib1_hybrid_height` into a patch for Iris-GRIB, since after the Iris-GRIB v0.15 release the `climatological` feature was released within Iris v2.3, breaking `test_grib1_hybrid_height`.

bjlittle:
@trexfeathers Okay, I'd say pop `pep8` into the `environment.yml` for now. 

Ultimately, we'll move `iris-grib` to be `black` compliant akin to `iris`, but until then it makes to explicitly add `pep8` as a testing dependency.

When we do move to `black` then we can remove the `pep8` dependency and the test.

Sound like a reasonable plan?

bjlittle:
Note that, the `iris-grib` test will start working from `iris` 2.4.* onwards.

bjlittle:
Awesome, thanks @trexfeathers :+1:

PR #185, opened 2020-01-16
equalise_attributes compatibility for Iris 2.4 and also master.

trexfeathers:
Should allow test_hybrid_coords_round_trip.py to pass both with the existing Iris and also future releases


pp-mo:
Thanks @trexfeathers 
Master tests fixed  👍 👍 👍 !

PR #187, opened 2020-01-31
Add support for Lambert Azimuthal Equal Area Projection

tpowellmeto:
Add support for loading GRIB2 template 3.140
https://apps.ecmwf.int/codes/grib/format/grib2/templates/3/140


tpowellmeto:
bump

tpowellmeto:
Hi guys, this one has been hanging around since Jan, is there anything extra that needs doing before this gets merged? This is required to load MO UKV GRIB files into Iris.

iris_grib/message.py
```diff
@@ -143,6 +143,9 @@ def data(self):
                 raise TranslationError(msg)
             if template in (20, 30, 90):
                 shape = (grid_section['Ny'], grid_section['Nx'])
+            elif template == 140:
+                shape = (grid_section['numberOfPointsAlongYAxis'],
+                         grid_section['numberOfPointsAlongXAxis'])
```
pp-mo, line 148: Have you checked that we can't just use "Nx" and "Ny", and so integrate this into the section above ?

iris_grib/_load_convert.py
```diff
@@ -928,10 +928,16 @@ def _calculate_proj_coords_from_grid_lengths(section, cs):
     mm_to_m = 1e-3
 
     if section['gridDefinitionTemplateNumber'] in _XYGRIDLENGTH_GDT_NUMBERS:
-        dx = section['Dx']
-        dy = section['Dy']
-        nx = section['Nx']
-        ny = section['Ny']
+        if section['gridDefinitionTemplateNumber'] == 140:
+            dx = section['xDirectionGridLengthInMillimetres']
```
pp-mo, line 932: Did you check that the shorter names don't in fact work?
 -- if not, we wouldn't need this test, or this section because the generic ("else") code can do it.

I ask because I can see that the template definition 3.140 in the official docs (https://community.wmo.int/activity-areas/wmo-codes/manual-codes/volume-i2)  _does_ at least reference the shorter names, 
e.g. "31–34  Nx – number of points along the x-axis".

In the end though, we can only use what implemented in eccodes.

pp-mo:
@tpowellmeto  Sorry for raising points after making you wait for ages
-- we are ready to get this done now !

Your test is very tidy, thanks.

pp-mo:
@tpowellmeto I'm just wondering if you also have some "real data" which you could release in public (in iris-test-data), so we can have a practical fallback when doing code development ? 
It would need to be something where a single message is reasonably small, say a few Mb at most.

Whereas a focussed unit-test is very useful for picking up odd mistakes and changes, "real" data is needed to support integration testing, which is invaluable if code needs refactoring.

I say this because we have had many problems in the past with GRIB compatibility
-- I think because the specs are so difficult to interpret precisely.
We write new templates, and then people find files for which they don't work !

iris_grib/message.py
tpowellmeto: noted, unfortunately when I tried this it didn't work hence adding these more complete names.

iris_grib/_load_convert.py
tpowellmeto: as above

tpowellmeto:
@pp-mo WRT the request for "real-data", unfortunately I don't have the authority off the bat to release operational data into the wild. I will take this as an action and follow up with Service Hub owners to see what is possible. So as not to block this change if I can get some I'll raise as a future PR.   :)

pp-mo:
Thanks @tpowellmeto  !

PR #189, opened 2020-02-19
Get iris-grib tests passing with the latest version of eccodes

lbdreyer:
Previously a couple tests were failing with eccodes> 2.14 as eccodes was no longer raising an error when you give it a mock object.  This fixes those tests.

The environment currently resolves with eccodes 2.16

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -2,5 +2,5 @@ channels:
     - conda-forge
 dependencies:
     - iris>=2
-    - python-eccodes>=0.9.1,<2
+    - python-eccodes
     - pep8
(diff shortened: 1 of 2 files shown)


environment.yml
```diff
@@ -2,5 +2,5 @@ channels:
     - conda-forge
 dependencies:
     - iris>=2
-    - python-eccodes>=0.9.1,<2
+    - python-eccodes
```
pp-mo: Does this mean they have fixed the version-going-backwards problem ??? :stuck_out_tongue:

PR #190, opened 2020-02-19
Fix dask2 proxies 

pp-mo:
Change that parallels https://github.com/SciTools/iris/pull/3659

**NOTE:** 
This contains a [temporary copy](https://github.com/SciTools/iris-grib/blob/fbe926137dafea6eac4db908d7f1c18f56f35002/iris_grib/__init__.py#L33) of the [key Iris routine](https://github.com/SciTools/iris/blob/9b8f9cbf943b2c659dcae2765e1cab089cabc645/lib/iris/util.py#L918) that enables this fix.
This can be removed when Iris 2.4 is out.
Not clear if we should merge this as-is, or wait for that ?

### DIFF
--- a/iris_grib/__init__.py
+++ b/iris_grib/__init__.py
@@ -20,79 +20,13 @@ import gribapi
 import numpy as np
 import numpy.ma as ma
 
+# NOTE: careful here, to avoid circular imports (as iris imports grib)
 import iris
 from iris._lazy_data import as_lazy_data
 import iris.coord_systems as coord_systems
 from iris.exceptions import TranslationError, NotYetImplementedError
+from iris.util import _array_slice_ifempty
 
-try:
-    from iris.util import _array_slice_ifempty
-except ImportError:
-    # A temporary cut-and-paste hack, until this is in an Iris release ...
-    # see : https://github.com/SciTools/iris/pull/3659
-    def _array_slice_ifempty(keys, shape, dtype):
-        """
-        Detect cases where an array slice will contain no data, as it contains
-        a zero-length dimension, and produce an equivalent result for those
-        cases.
-
-        The function indicates 'empty' slicing cases, by returning an array
-        equal to the slice result in those cases.
-
-        Args:
-
-        * keys (indexing key, or tuple of keys):
-            The argument from an array __getitem__ call.
-            Only tuples of integers and slices are supported, in particular no
-            newaxis, ellipsis or array keys.
-            These are the types of array access usage we expect from Dask.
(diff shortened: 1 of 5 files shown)


iris_grib/__init__.py
```diff
@@ -25,6 +25,73 @@
 import iris.coord_systems as coord_systems
 from iris.exceptions import TranslationError, NotYetImplementedError
 
+try:
+    from iris.util import _array_slice_ifempty
```
lbdreyer: I'm pretty uncomfortable about this using a private function from iris.

I don't have a better suggestion, but I do think we seriously need to consider how we will manage this sort of thing, as we will encounter this problem more and more as we try to break iris up

pp-mo: 🤔 I really see what you mean.  I think this is a big problem for the future.
Some thoughts ...

  * We may want to share common functionality with multiple dependent projects , but retain some ability to change interface and/or behaviour without it forcing a major version increment.

  * *For example*, I can imagine we might want common code for things like : a file-loader plugin class; a common DataProxy concept; DataManager objects;, 'lazy' data functions.

  * The key problem is  :  At times we will want to change such things, in ways which will force updates in the dependent projects.  ***But,*** we don't want this to have the same implications as breaking changes in user-api, and it certainly isn't appropriate to apply SemVer rules in the same way.

  * The key driver of this is that projects like iris-grib (e.g. file format interfaces) want to share common functions, but these may be lower-level and not user-relevant.  It is a flexibility/stability tradeoff.  There is no benefit in making all shared function rigidly defined + controlled in the same way as user API : it needs to be more flexible than that.

  * we _could_ put such functions in their own independent projects (not just in Iris), but I don't think that actually adds anything useful.

I think it's also worth noting that, in the specific case of this function, **not only** is it too low-level, without user-level relevance, to publish it as a controlled API, **but also** it is highly specific to what Dask is doing, and that might even change, which makes it a bit "ad hoc".

pp-mo:
Thanks @lbdreyer 
> removing the try/except ...  iris >= 2.4 ... depend on the Iris release

I think it's close.  Let's wait ...

pp-mo:
> removing the try/except ... iris >= 2.4 ... depend on the Iris release

Got the release, but now we need a conda-forge recipe update ...

pp-mo:
> a conda-forge recipe update

Done + ready !
Are we go, @lbdreyer ?

pp-mo:
Had to rebase.
Ok now, :point_right:  @lbdreyer ?

Issue #56, opened 2016-08-12
Fix GRIB1 support


pp-mo:
The GRIB1 loading translation are still using the old "GribWrapper" code, which is the only thing preventing us throwing this old code away.

There is still a slippery kludge that means when you load a GRIB1 message the 'field' argument in a callback is a GribWrapper instead of a GribMessage ([here](https://github.com/SciTools/iris-grib/blob/d81db4de9977a39de9f2d8996f3638ee27600792/iris_grib/__init__.py#L697)).

So, this is now the **only** thing that GribWrapper is still used for, and the only remaining occasion that a user might possibly be concerned about what a GribWrapper looks like.

We haven't addressed this yet because GRIB1 usage is very low.

We should rewrite the Grib1 loader to use the modern message object + throw away all the old GribWrapper-based processing.

PR #191, opened 2020-02-19
Port test_grib_load_translations from iris

pp-mo:
Tests ported from `iris.tests.test_grib_load_translations`.
Renamed and slightly simplified, as this (now) **only** tests the old GRIB-1 mechanisms.

A case could be made for combining this with the existing "sister" testcode `iris_grib.tests.unit.grib1_load_rules.test_grib1_convert`, and unifying the approaches.
But for GRIB1 I think this is strictly not worth it.  
Maybe when we [port GRIB1 to the new "strict" loader code](https://github.com/SciTools/iris-grib/issues/56) ?

### DIFF
--- a/iris_grib/tests/unit/grib1_load_rules/test_grib1_load_translations.py
+++ b/iris_grib/tests/unit/grib1_load_rules/test_grib1_load_translations.py
@@ -1,6 +1,6 @@
-# Copyright Iris contributors
+# Copyright iris-grib contributors
 #
-# This file is part of Iris and is released under the LGPL license.
+# This file is part of iris-grib and is released under the LGPL license.
 # See COPYING and COPYING.LESSER in the root of the repository for full
 # licensing details.
 """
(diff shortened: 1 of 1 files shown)


lbdreyer:
> But for GRIB1 I think this is strictly not worth it.

I definitely agree with this!

iris_grib/tests/unit/grib1_load_rules/test_grib1_load_translations.py
```diff
@@ -0,0 +1,384 @@
...
+                "timeRangeIndicator": 1,
+                "P1": 2,
+                "P2": 0,
+                # time unit - needed AS WELL as 'indicatorOfUnitOfTimeRange'
+                "unitOfTime": 1,
+                "table2Version": 9999,
+            }
+        )
+        # Add edition-dependent settings.
+        self["edition"] = edition
```
lbdreyer: Arguably you could edit this... but I don't think we need to worry about that right now

lbdreyer: Actually I think I've misunderstood the purpose of this, so ignore that last comment

lbdreyer:
Overall I'm happy with this. I'll merge...

PR #192, opened 2020-02-20
Port tests from iris/tests/integration/test_grib2.py

lbdreyer:
Companion Iris PR: https://github.com/SciTools/iris/pull/3664


lbdreyer:
GRIB saving with eccodes 2.16 seems to be broken as is causing the tests to fail.

I suggest we pin back to eccodes 2.14 #193.
Once that's in, these tests should pass

iris_grib/tests/integration/round_trip/test_grid_definition_section.py
```diff
@@ -0,0 +1,140 @@
...
+    def test_save_load(self):
+        # Load sample UKV data (variable-resolution rotated grid).
+        path = tests.get_data_path(("PP", "ukV1", "ukVpmslont.pp"))
+        cube = load_cube(path)
+
+        # Extract a single 2D field, for simplicity.
+        self.assertEqual(cube.ndim, 3)
+        self.assertEqual(cube.coord_dims("time"), (0,))
+        cube = cube[0]
+
```
lbdreyer: This cube loads in with a shape of (928, 744).

6 years ago, eccodes had a bug such that it could only save square data. Thus, in Iris this tests square off the data before saving it (see [here](https://github.com/SciTools/iris/blob/836244ea92d4dff32e411be329524234a5f9f6dd/lib/iris/tests/integration/test_grib2.py#L180)). I have removed that as the bug has been fixed (many years ago!)

pp-mo:
Thanks @lbdreyer : Very conscientious to have redistributed these so carefully !
:+1: :1st_place_medal: 
I'm satisfied it's all good here, but let's see if #193 will get us a green tick ...

PR #193, opened 2020-02-20
Pin eccodes to avoid bad version

lbdreyer:
Unfortunately, due to problems with saving in the latest eccodes version (2.16) (as shown [in these tests](https://travis-ci.org/SciTools/iris-grib/jobs/652968694?utm_medium=notification&utm_source=github_status)), we will have to pin back to using an older version of eccodes.

This is effectively undoing the changes in #189

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -2,5 +2,5 @@ channels:
     - conda-forge
 dependencies:
     - iris>=2
-    - python-eccodes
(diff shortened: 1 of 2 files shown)


pp-mo:
> This is effectively undoing the changes in #189

yes :-(

PR #196, opened 2020-02-20
Port tests from iris/tests/integration/test_pickle.py

lbdreyer:
Companion PR: https://github.com/SciTools/iris/pull/3667

Note this will probably cause conflicts with #194 but I will resolve those once either of these go in.

I wanted to get this up to see what you thought of this approach


lbdreyer:
I think this requires some more thinking. I'll try putting something else up

PR #198, opened 2020-02-21
Port tests from iris/tests/integration/test_grib_save.py

lbdreyer:
Companion PR: https://github.com/SciTools/iris/pull/3669

Basically a lift and shift of the tests in iris/tests/integration/test_grib_save.py other than the minor necessary changes such as inherit from `iris_grib.tests.IrisGribTest` rather than `iris.tests.IrisTest`

I just moved it to a file called iris_grib/tests/integration/save_rules/test_grib_save.py  as I thought that would be the easiest to review, but I could move it somewhere more sensible, iris_grib/tests/integration/save_rules/__init__.py for example.

### DIFF
--- a/iris_grib/tests/integration/save_rules/test_grib_save.py
+++ b/iris_grib/tests/integration/save_rules/test_grib_save.py
@@ -1,0 +1,326 @@
+# Copyright iris-grib contributors
+#
+# This file is part of iris-grib and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
+"""
+Tests for specific implementation aspects of the grib savers.
+Ported here from 'iris.tests.integration.test_grib_save'.
+
+"""
+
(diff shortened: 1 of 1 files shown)


iris_grib/tests/integration/save_rules/test_grib_save.py
```diff
@@ -0,0 +1,326 @@
...
+                    ),
+                    1949.0,
+                )
+
+
+class TestHandmade(tests.IrisGribTest):
+    def _lat_lon_cube_no_time(self):
+        """
+        Returns a cube with a latitude and longitude suitable for testing
+        saving to PP/NetCDF etc.
```
lbdreyer: I suppose this no longer makes sense in the context of testing for GRIB. 
I'll change this docstring

lbdreyer:
I've now updated this PR with an extra commit that edit the tests slightly to make them more understandable in the context of iris_grib

PR #200, opened 2020-02-21
Import tests from iris.test.integration.test_grib load

pp-mo:
Removed from iris : https://github.com/SciTools/iris/pull/3670

### DIFF
--- a/iris_grib/tests/integration/load_convert/test_sample_file_loads.py
+++ b/iris_grib/tests/integration/load_convert/test_sample_file_loads.py
@@ -65,13 +65,6 @@ class TestBasicLoad(tests.IrisGribTest):
         self.assertCML(cubes,
                        _RESULTDIR_PREFIX + ("missing_values_grib2.cml",))
 
-    @skip_irisgrib_fails
-    def test_y_fastest(self):
-        cubes = iris.load(
-            tests.get_data_path(("GRIB", "y_fastest", "y_fast.grib2"))
-        )
-        self.assertCML(cubes, _RESULTDIR_PREFIX + ("y_fastest.cml",))
-
     def test_polar_stereo_grib1(self):
         cube = iris.load_cube(
             tests.get_data_path(("GRIB", "polar_stereo", "ST4.2013052210.01h"))
--- a/iris_grib/tests/integration/load_convert/test_sample_file_loads.py
+++ b/iris_grib/tests/integration/load_convert/test_sample_file_loads.py
@@ -62,7 +62,8 @@ class TestBasicLoad(tests.IrisGribTest):
             ("GRIB", "missing_values", "missing_values.grib2")
         )
         cubes = iris.load(gribfile)
-        self.assertCML(cubes, _RESULTDIR_PREFIX + ("missing_values_grib2.cml",))
+        self.assertCML(cubes,
+                       _RESULTDIR_PREFIX + ("missing_values_grib2.cml",))
 
     @skip_irisgrib_fails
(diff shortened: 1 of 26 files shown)


iris_grib/tests/integration/load_convert/test_sample_file_loads.py
```diff
@@ -0,0 +1,223 @@
+# Copyright iris-grib contributors
+#
+# This file is part of iris-grib and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
+"""
+Integration tests for grib2 file loading.
+
+These load various files from iris-test-data, and compare the cube with a
```
lbdreyer:
```suggestion
These tests load various files from iris-test-data, and compare the cube with a
```

A minor change but I found this sentence a bit hard to read. I think it's not clear what you're referring to by only having "These" as the object of the sentence, especially when it is the start of a new paragraph.

iris_grib/tests/integration/load_convert/test_sample_file_loads.py
```diff
@@ -0,0 +1,223 @@
...
+    def test_load_masked(self):
+        gribfile = tests.get_data_path(
+            ("GRIB", "missing_values", "missing_values.grib2")
+        )
+        cubes = iris.load(gribfile)
+        self.assertCML(cubes,
+                       _RESULTDIR_PREFIX + ("missing_values_grib2.cml",))
+
+    @skip_irisgrib_fails
+    def test_y_fastest(self):
```
lbdreyer: Why do you want to keep this test? We had already removed it from the iris repo: https://github.com/SciTools/iris/pull/3632

iris_grib/tests/integration/load_convert/test_sample_file_loads.py
```diff
@@ -0,0 +1,223 @@
...
+# loader does not support X,Y dimensioned data (as opposed to Y,X).
+# TODO: fix this, or remove the test.
+skip_irisgrib_fails = skipIf(
+    True, "Current grib2 loader does not support this test(s)."
+)
+
+_RESULTDIR_PREFIX = ("integration", "load_convert", "sample_file_loads")
+
+
+@tests.skip_data
```
lbdreyer: We don't typically do this for iris-grib tests as, unlike iris, we *always* download the iris-test data.
That said, maybe it's a good idea to future-proof these tests, should we decide to copy what iris does.

iris_grib/tests/integration/load_convert/test_sample_file_loads.py
pp-mo: Ok, I hadn't spotted that when I did this.

iris_grib/tests/integration/load_convert/test_sample_file_loads.py
pp-mo: I did check for usage within iris-grib and there are lots of existing uses, so I thought it best to keep it.

iris_grib/tests/integration/load_convert/test_sample_file_loads.py
pp-mo: done.

iris_grib/tests/integration/load_convert/test_sample_file_loads.py
lbdreyer: I see you are right, there are some that do use test.skip_data, but not all. We should try to be consistent with this and perhaps always include the `@test.skip_data`

PR #205, opened 2020-03-03
Depth scaling

BHFock:
Scale 'depth' coordinates to cm or mm to handle cm / mm values for soil levels when writing to GRIB2


trexfeathers:
@BHFock we're very sorry that this PR was lost for so long - there has been some lack of clarity about who looks after iris-grib in the past. We expect that you are working on other things these days, and we are also interested in creating a global solution that will address this and several other issues via common code.

Therefore:

- We have raised #538
- We are closing this pull request

We are keen to show better ownership to iris-grib, so if you or anyone else is keen to see this problem solved, please hold us to account about finishing #538 promptly.

PR #206, opened 2020-05-29
Add support for missing timeIncrement in template 4.8

m1dr:
Add support for loading GRIB2 files using template 4.8 in which the time increment between successive fields used in statistical processing is not defined.


pp-mo:
Thanks @m1dr looks fine !

PR #210, opened 2020-06-15
Correct rounding of vertical coordinates.

pp-mo:
IMHO we should be rounding ***any*** time we convert a float value to a scaled integer.
See: #211
However, it looks like that is awkward and will get in the way of the 0.16 release, so I think it's not worth it now.


lbdreyer:
This looks sensible to me, more safe than just using int!

I assigned myself yesterday as I wasn't sure anyone was looking at this, but it sounds like @alastair-gemmell may be? If so I'll hold off on merging.

PR #213, opened 2020-06-17
Tests for saving other-format data into grib : ported from Iris.

pp-mo:
These are ported directly across from [these iris integration tests](https://github.com/SciTools/iris/tree/be9e2ff703e4a87f2c2533e86bd06aa402563b80/lib/iris/tests/integration/format_interop).

The aim of taking these tests out of Iris is so that we can avoid ***any*** need for iris-grib within Iris' own tests.
(not *quite* there yet, but coming soon...)

That linkage is a headache because at present *each requires the other* for testing :  Therefore, any breaking change (that requires parallel changes to both) can prevent both from passing tests, until both cut new releases.

----
NOTE: migrating these tests here still leaves open a possible ***future*** dependency problem ...
If in future we "split out" either of these other file-formats (name + pp) to a separate package, as we have done with iris-grib itself, that will make iris-grib testing depend on those other file-format packages.
I propose that we simply leave that as a "problem for the future".

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -2,5 +2,6 @@ channels:
     - conda-forge
 dependencies:
     - iris>=2.4
+    - mo_pack
     - python-eccodes
     - pep8
--- a/iris_grib/tests/integration/format_interop/__init__.py
+++ b/iris_grib/tests/integration/format_interop/__init__.py
@@ -1,0 +1,12 @@
+# Copyright iris-grib contributors
+#
+# This file is part of iris-grib and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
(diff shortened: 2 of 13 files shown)


pp-mo:
> a possible future dependency problem ... depend on those other file-format packages.

Well, the need for mo-pack here shows exactly what that meant !

trexfeathers:
> NOTE: migrating these tests here still leaves open a possible future dependency problem ...
If in future we "split out" either of these other file-formats (name + pp) to a separate package, as we have done with iris-grib itself, that will make iris-grib testing depend on those other file-format packages.
I propose that we simply leave that as a "problem for the future".

😨 😨

PR #214, opened 2020-06-24
Fix iris3 coord units

### LINKED ISSUE #207: Check behaviour with Iris 3.0-style Coord changes
Especially because Coord units now default to "unknown" : Check that results from grib loading are always sensible (mostly, should be =unchanged)
See https://github.com/SciTools/iris/issues/3721

pp-mo:
**WIP: DO NOT MERGE**
until we have Iris3, and this is re-tested + OK

Closes #207

These changes will be needed when we release the changes on iris [default_units](https://github.com/SciTools/iris/tree/default_units)  feature-branch 
(but which is not yet merged, hence WIP flag). 
They just ensure "units=1", for unchanged behaviour when creating coords.
That is anyway benign :  it all works with current Iris.

( checked out locally, against the Iris feature-branch current content )

### DIFF
--- a/iris_grib/_grib1_load_rules.py
+++ b/iris_grib/_grib1_load_rules.py
@@ -238,9 +238,9 @@ def grib1_convert(grib):
     if \
             (grib.levelType == 'ml') and \
             (hasattr(grib, 'pv')):
-        aux_coords_and_dims.append((AuxCoord(grib.level, standard_name='model_level_number', attributes={'positive': 'up'}), None))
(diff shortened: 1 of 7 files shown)


pp-mo:
Added 2 new commits to fix for recent changes in Iris : Notably from https://github.com/SciTools/iris/pull/3804, which has rationalised types of coordinate-system properties (most are now cast to floats)

**This tested all passed with [this recent Iris commit](https://github.com/SciTools/iris/tree/84dc829d3b7d37dc2078266f1a92cf57b14d89ab)**

pp-mo:
Now done in #231

PR #224, opened 2020-09-29
Remove fix

pp-mo:
**WIP**
Aiming to remove a workaround introduced in #208, supposedly now [fixed in python-eccodes](https://github.com/SciTools/iris-grib/pull/208#issuecomment-667335469).

So far, have re-checked that the original problem is replicable, having forced the old python-eccodes version, and removed the fix.  This triggered a segfault in the `tests.integration.round_trip.test_grid_definition_section.TestGDT5`.  It *should* also hit this in  `tests/unit/save_rules/test_data_section.TestNonDoubleData` (but that is not run before the crash happens).
**Then**, removed the version pin, now testing with 2.18.0 and python-eccodes 2020.06.0.
This should have been ok, but at present the original crash is still occurring .

**Current Status** : skipped out the GDT5 test, temporarily, to enable the intended test `tests/unit/save_rules/test_data_section.TestNonDoubleData`, which targets/demonstrates the problem more clearly.

### DIFF
--- a/iris_grib/_save_rules.py
+++ b/iris_grib/_save_rules.py
@@ -1490,12 +1490,6 @@ def data_section(cube, grib):
         gribapi.grib_set(grib, "bitmapPresent", 1)
         gribapi.grib_set_double(grib, "missingValue", fill_value)
 
-    # A segmentation fault is raised by `gribapi.grib_set_double_array` if it
-    # tries to cast large data to float64. As a temporary fix we cast the data
-    # upfront
-    # TODO: remove the `astype` command once eccodes (gribapi) has been fixed.
-    if data.dtype != np.float64:
-        data = data.astype(np.float64)
     gribapi.grib_set_double_array(grib, "values", data.flatten())
 
     # todo: check packing accuracy?


pp-mo:
Ok. I think that shows the same bug, according to [discussion in #208](https://github.com/SciTools/iris-grib/pull/208#issue-425683808)

pp-mo:
Oh dear.
It looks like removal of this workaround will have to wait
:-(

trexfeathers (commented):
😥

lbdreyer:
It has been fixed in pytho-eccodes but not been released on conda yet, only pip.

pp-mo:
> not been released on conda yet, only pip
Well I'm a bit confused by that. The releases available via conda seem to be long after the merge of the referenced issue ?
Anyway, I raised https://github.com/ecmwf/eccodes-python/issues/34.  Maybe will get a clarifying response

lbdreyer:
> The releases available via conda seem to be long after the merge of the referenced issue ?

The fix (https://github.com/ecmwf/eccodes-python/pull/31) went in at the end of July. The last conda release of python eccodes was at the end of June, THe last pip release was in August. https://github.com/ecmwf/eccodes-python/releases
So it's not available on conda.
I asked when it will be available: https://github.com/ecmwf/eccodes-python/issues/33

shahramn:
The segmentation fault issue with float32 is fixed in the next release of eccodes-python (v1.0.0) which will be accompanied by ecCodes v2.19.0
Hopefully in October

pp-mo:
Re-fixed and re-targetted at v0.16.x

It's now _really_ simple, so let's hope it checks out..

lbdreyer:
Thanks @pp-mo !

PR #226, opened 2020-09-30
Requirements re-haul and optional testing against Iris latest

trexfeathers:
This PR:

- Replicates the requirements structure of Iris introduced in SciTools/iris#3812
- Re-introduces the **option** of testing against the latest iris source code, which was previously removed in #100
  The Travis matrix now tests against the latest release, and the latest source code

As discussed offline with @pp-mo, my conflicting time commitments have stopped me testing this, so we have agreed that testing can be part of a more thorough review process before it's appropriate to be merged.


trexfeathers:
@pp-mo not planning to do any more work on this. Please take a look and see how we can get the tests working!

I also haven't looked too closely about whether the pip set up is OK. One question that springs to mind is whether `wheel` should be part of requirements/setup.txt.

pp-mo:
Replaced by #231 : on new feature-branch
( Most of the code there is actually derived from this. )

PR #229, opened 2020-10-05
Adopt newstyle dependencies; fix travis to suit.

pp-mo:
This is basically the content of #226 , which it aims to replace. 
except ...
  * rebased onto latest master
  * fixed .travis.yml to get it all working (I hope)
  * removed (for now) the code to test against latest-Iris instead of latest-release : *just to simplify the PR*

You can compare the .travis.yml directly with the one from #226 
 : https://github.com/trexfeathers/iris-grib/compare/iris-latest...pp-mo:new_deps#diff-354f30a63fb0907d4ad57269548329e3

The **_whole_** of this doesn't compare well with #226, owing to changes since.
But the things that this PR affects are intelligible, essentially unchanged from #226 and can be compared to master (i.e. in "files changed")


pp-mo:
When we have this agreed + merged, I will re-introduce the test-against-Iris-master code.
We will get that running + testing, but possibly in a feature-branch as to work it needs test changes which will then **_break_** testing with the old Iris release.

requirements/ci/py36.yml
```diff
@@ -0,0 +1,22 @@
...
+
+channels:
+  - conda-forge
+
+dependencies:
+  - python=3.6
+
+# Setup dependencies.
+  - setuptools
+
```
trexfeathers, line 11: Is it appropriate to include `wheel` in here too?

requirements/ci/py37.yml
```diff
@@ -0,0 +1,22 @@
...
+
+channels:
+  - conda-forge
+
+dependencies:
+  - python=3.7
+
+# Setup dependencies.
+  - setuptools
+
```
trexfeathers, line 11: Is it appropriate to include `wheel` in here too?

requirements/setup.txt
```diff
@@ -0,0 +1,4 @@
+# Dependencies necessary to run setup.py of iris-grib
+# ---------------------------------------------------
+
+setuptools
```
trexfeathers, line 4: Is it appropriate to include `wheel` in here too?

trexfeathers (commented):
Looks good @pp-mo

Just one question outstanding from my own PR

pp-mo:
Re:
> Is it appropriate to include wheel in here too?

In my understanding, the wheel package is only required if you want to _build_ wheels.
cf. https://github.com/pypa/wheel

So, I **_think_** this would only be needed if you need to do a "setup.py bdist_wheel",
thus, no reason to require it here.

trexfeathers (approved):

PR #231, opened 2020-10-05
Travis test with  both Iris latest-release and latest-master.

pp-mo:
Targetting the "f__tes-iris-master" feature branch.

This is intended to develop the feature-branch for testing against latest iris, with a staged development, tested at each commit.
Plan :
  1. [x] test against Iris both old (2.4) and new (3.0.0.rc0)
    * it will fail when testing the 3.0rc
  2. [x] add fixes to tests for Iris 3
    * it will succeed against Iris3, and now fail against 2.4
  3. [x] restrict to testing against latest-master only (\*\*)
    * passes : **mergable** to the feature-branch
  
(\*\*) : at point #3, it may also be appropriate to require Iris >= 3

After this is merged, we will wait until the new Iris 3 is released on conda-forge.
Then we can mergeback to master and tag the 0.16 release.


trexfeathers:
It hadn't occurred to me that this feature branch might want to test against released Iris, too. What's the scenario in which that's needed?

pp-mo:
> test against released Iris, too

You beat me to it !
Changes to come show that it's not practical to make it test against both, as some tests would need to allow either of two CML results.  Hence, I intend to restrict to latest-Iris only at that point.

pp-mo:
*Now* : works with olf Iris but not new.
Coming next : fixes that make it work with new, but not old ...

pp-mo:
**Now** : tests all pass with new-Iris (3), fails with old (2.4)
next : restrict to testing against "new" only.

pp-mo:
Ok, shazamm !! :mage_man: :sparkles: 
Please review, and we'll merge to the f-b when happy.

iris_grib/tests/unit/grib1_load_rules/test_grib1_convert.py
```diff
@@ -116,20 +116,17 @@ def test_grib1_hybrid_height(self):
...
-
-        aux_coord_defns = [coord._as_defn() for coord, dim in results[8]]
-        self.assertIn(ml_ref, aux_coord_defns)
-        self.assertIn(lp_ref, aux_coord_defns)
-        self.assertIn(s_ref, aux_coord_defns)
+        coords_and_dims = results[8]
+        coord, = [co for co, _ in coords_and_dims
+                  if co.name() == 'model_level_number']
+        self.assertEqual(coord.units, '1')
+        self.assertEqual(coord.attributes['positive'], 'up')
```
trexfeathers, line 123: Should you also assert empty `attributes` for the other two `coord`s?

iris_grib/tests/unit/grib1_load_rules/test_grib1_convert.py
```diff
@@ -116,20 +116,17 @@ def test_grib1_hybrid_height(self):
...
-                                       {}, None, False)
-        s_ref = iris.coords.CoordDefn(None, 'sigma', None,
-                                      cf_units.Unit('1'),
-                                      {}, None, False)
-
-        aux_coord_defns = [coord._as_defn() for coord, dim in results[8]]
-        self.assertIn(ml_ref, aux_coord_defns)
-        self.assertIn(lp_ref, aux_coord_defns)
-        self.assertIn(s_ref, aux_coord_defns)
+        coords_and_dims = results[8]
```
trexfeathers, line 119: Your avoidance of defining coordinates means:

* The `cf_units` import is unused
* You could cut the `iris` import down to just importing `HybridPressureFactory`, if you're happy with that style

trexfeathers (commented):
Looks great @pp-mo, thanks for the clear commit sequence too.

Couple of questions on good testing practice...

trexfeathers:
I had imagined this feature branch persisting for future use, with appropriate mergebacks being made to `master` when needed. Is that the intention?

pp-mo:
> feature branch persisting

I think master should be testing against a released Iris.
So we can merge back once there is a released Iris we can install from conda-forge, but not before.
( Also, I would aim to leave the matrix option controls in place, commented out, so we have banked the solution for next time. )

iris_grib/tests/unit/grib1_load_rules/test_grib1_convert.py
pp-mo: Well, I thought it was OK to just check the things which the convert routine produces.
My reading of the previous code is that it **_has_** to provide all args to create a CoordDefn, but it's not really  important to check every aspect.

trexfeathers:
> Also, I would aim to leave the matrix option controls in place, commented out, so we have banked the solution for next time.

OK cool, so it'll be easy to make a new feature branch if there's something specific that needs testing in the future, but we won't need to persist the current feature branch in the meantime.

pp-mo:
> we won't need to persist the current feature branch

Ah yes, I see -- I guess that is what I was imagining.

Thanks @trexfeathers !

PR #234, opened 2020-10-12
Cosmetic change : rename the travis iris-test-version options

pp-mo:
Rename Iris test-version options, and enable all to check action.

This will also demonstrate testing against latest release
  -- but then some tests will fail

When we get to merge this F-B back, it is because we have an actual Iris 3.0 release.
In that case, this should start to work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,10 @@ dist: xenial
 
 env:
   matrix:
-    - PYTHON_VERSION=36 LIMIT_TO_RELEASED_IRIS=false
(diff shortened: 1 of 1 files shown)


trexfeathers (changes requested):
I'm very much in favour of something less cumbersome than `LIMIT_TO_RELEASED_IRIS`, but I'm not a fan of `VERSION`, which already has another, different meaning. `IRIS_SOURCE` perhaps? Or something else if you can think of it.

trexfeathers (approved):

PR #235, opened 2020-10-12
Fix RC date in release notes (about to cut).

pp-mo:
Tiny fix (sorry about this).
When done, I will cut the actual release tag.


trexfeathers:
Looks good. I hadn't noticed that "What's New"s state the eventual version even when it's still an RC, but that's definitely the case for Iris as well 👍

trexfeathers (approved):

PR #240, opened 2021-01-19
Add 'main' conda-forge channel, needed for docs builds.

pp-mo:
This fixes the RTD docs-build error which emerged following the cut of the original 0.16.0rc0 pre-release tag.

Given the error seen in the attempt to re-spin #224 today, we suspect those errors will come through here too
(something to do with ellipsoid definitions changing)
So we will probably want to wait until that is fixed before merging this ...

### DIFF
--- a/iris_grib/__init__.py
+++ b/iris_grib/__init__.py
@@ -342,55 +342,22 @@ class GribWrapper:
         self.extra_keys['_forecastTimeUnit'] = self._timeunit_string()
 
         # shape of the earth
-
-        # pre-defined sphere
-        if self.shapeOfTheEarth == 0:
+        soe_code = self.shapeOfTheEarth
+        # As this class is now *only* for GRIB1, 'shapeOfTheEarth' is not a value read from the actual file :  It is
+        # really a GRIB2 param, and the value is merely what eccodes (griabpi) considers the correct default.
+        # This was always =6 until eccodes 0.19, when it changed to 0.  See https://jira.ecmwf.int/browse/ECC-811
+        # The two represent different sized spherical earths.
+        if soe_code not in (6, 0):
+            raise ValueError('Unexpected shapeOfTheEarth value =', soe_code)
+        # *FOR NOW* maintain the old behaviour (radius=6371229) in all cases for backwards compatibility.
+        # However, this does not match 'radiusOfTheEarth' and may be incorrect :  We may change it in future.
+        soe_code = 6
+        if soe_code == 0:
+            # New supposedly-correct correct value, matches the 'radiusOfTheEarth' parameter.
             geoid = coord_systems.GeogCS(semi_major_axis=6367470)
-
-        # custom sphere
-        elif self.shapeOfTheEarth == 1:
-            geoid = coord_systems.GeogCS(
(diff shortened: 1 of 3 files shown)


pp-mo:
Ok, so the key problem is that ECMWF have changed a default parameter setting for GRIB1 messages.
I'm not sure quite how this works, but it looks like 'shapeOfTheEarth' used to be defined as '6' but is now '0'.

All the problems are with GRIB1 files in iris-test-data, that now read back differently, 
for example `<testdata...>/GRIB/rotated_uk/uk_wrongparam.grib1`
That file has no section 3, but the shapeOfTheEarth parameter is a section-3 thing, so I think it is about defaults.

From comparing release downloads of eccodes, the key change seems to be in a file `definitions/grib1/section.2.def`
At version 1.18 this reads...
```
# START grib1::section
# SECTION 2, Grid description section
#  Length of section

position offsetSection2;
 . . .
transient shapeOfTheEarth=6 : hidden;
```
Whereas in 1.19 this last line changed to ...
`transient shapeOfTheEarth=0: hidden; #ECC-811`

In the [release notes](https://confluence.ecmwf.int/display/ECC/ecCodes+version+2.19.0+released) , Bug Fixes section, we have 
"[ECC-811] - shapeOfTheEarth=6 for GRIB1"
Unfortunately the issue content is private + you can't view it

pp-mo:
I've posted a 'fix' to that gribapi change.
From discussion with @lbdreyer, our main priority right now is backwards-compatibility.
We can discuss later [whether this was always really wrong and should in future be fixed](https://github.com/SciTools/iris-grib/issues/241).

iris_grib/__init__.py
```diff
@@ -342,56 +342,23 @@ def _compute_extra_keys(self):
...
+        # really a GRIB2 param, and the value is merely what eccodes (griabpi) considers the correct default.
+        # This was always =6 until eccodes 0.19, when it changed to 0.  See https://jira.ecmwf.int/browse/ECC-811
+        # The two represent different sized spherical earths.
+        if soe_code not in (6, 0):
+            raise ValueError('Unexpected shapeOfTheEarth value =', soe_code)
+        # *FOR NOW* maintain the old behaviour (radius=6371229) in all cases for backwards compatibility.
+        # However, this does not match 'radiusOfTheEarth' and may be incorrect :  We may change it in future.
+        soe_code = 6
+        if soe_code == 0:
+            # New supposedly-correct correct value, matches the 'radiusOfTheEarth' parameter.
```
lbdreyer, line 356: Is this intentional, you are hardcoding `soe_code` to be 6, but including the `if soe_code == 0` for future, if we decide to change it?

pp-mo: Yes, I thought I should remove most of the previous unused options, but I left both these ones in for reference.

lbdreyer:
Looks good to me!
As @pp-mo says I think we need to prioritise minimal change for this release.
We would first need to understand the impact of changing to the new default (covered by https://github.com/SciTools/iris-grib/issues/241), i.e. how many users would it upset as they expect the old, incorrect default,  and as we're pressed for time right now, I don't think this is the right time.

PR #246, opened 2021-02-07
add missing test dependency

bjlittle:
This PR adds a missing test dependency that is expected by `iris` when testing.

Typically a developer will already have this available when testing against a development version of `iris`, but the `requests` package is not pulled in by `iris-grib` when installing `iris` from `conda-forge`.


bjlittle:
Closed by #250

PR #249, opened 2021-02-24
Merge back v0.16.0

jamesp:
Merge back v0.16.0.  Version bumped to 0.17.dev0.

This PR replaces #248

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,10 @@ dist: xenial
 
 env:
   matrix:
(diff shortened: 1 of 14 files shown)


lbdreyer:
Looks good to me!
Thanks @jamesp !

lbdreyer:
We currently are not able to merge something without squashing the commits. That would mean we would lose the history of these commits, which I'm not keen to do!
I'm going to see if we can temporarily enable non-squashing merge

PR #271, opened 2021-06-02
pdt6 & gdt140

marqh:
This PR targets new GRIB2 interpretation functionality, specifically:
* GDT 3.140 (Lambert Azimuthal)
    * save (load already exists) 
* PDT 4.6 (percentile at a point in time)
    * load and save 

There is no standard name in CF for `percentile` which seems like an omission to me, so i have used a long name.

I may like to propose a standard name of `percentile` to CF, which would then facilitate a change to the test code for PDT 4.6, but it seems challenging to wait for this, and significant datasets are being produced with `long_name=='percentile'`. I would also expect iris-grib to be backwards compatible with the use of long name, which I think is delivered by the implementation of the `coords` method (_save_rules#L1308)

### DIFF
--- a/iris_grib/_load_convert.py
+++ b/iris_grib/_load_convert.py
@@ -2066,6 +2066,37 @@ def product_definition_template_1(section, metadata, frt_coord):
     metadata['aux_coords_and_dims'].append((realization, None))
 
 
+def product_definition_template_6(section, metadata, frt_coord):
+    """
+    Translate template representing percentile forecast,
+    at a horizontal level or in a horizontal layer at a
+    point in time.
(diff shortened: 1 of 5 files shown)


marqh:
hmm, looks like the test failure is due to blocking that may already be fixed by
#270 

perhaps someone could update me when #270 is merged so I can rebase and rerun tests?
ty

pp-mo:
Hi @marqh 
Just stumbled on this again.
given #270 resolved, could you now try rebasing ?

trexfeathers:
Replaced by #295.

PR #272, opened 2021-06-03
Prep for 0.17.1 release

lbdreyer:
This PR:
* updates the version `0.17.0` -> `0.17.1`
* updates the release notes (see built copy of docs here: https://iris-grib.readthedocs.io/en/docstest_0v17_1/)
* updates the iris requirement from `>=3` to `>=3.0.2`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@ Getting started
 2. Fork the iris-grib repository, create your new fix/feature branch, and
    start commiting code.
  - The main
-   [Iris development guide](http://scitools.org.uk/iris/docs/latest/developers_guide/gitwash/git_development.html)
+   [Iris development guide](https://scitools-iris.readthedocs.io/en/latest/developers_guide/gitwash/index.html)
    has more detail.
 3. Remember to add appropriate documentation and tests to supplement any new or changed functionality.
 
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,7 @@
 iris-grib
 =========
 
-|CirrusCI|_ |Coveralls|_
+|CirrusCI|_ |Coveralls|_ |conda-forge|_ |pypi|_ |License|_ |Commits|_ |Contributors|_
 
 GRIB interface for `Iris <https://github.com/SciTools/iris>`_.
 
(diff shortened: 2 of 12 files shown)


lbdreyer:
The PY=36 tests failed with: 
```
FileNotFoundError: [Errno 2] No such file or directory: '.nox/tests-iris-source/tmp/iris/requirements/ci/py36.yml'
nox > Session tests(iris='source') failed.
```
which doesn't make sense to me. I'm going to rerun them in case that magically fixes the problem

trexfeathers:
>The PY=36 tests failed with...

Because SciTools/iris#4163 was merged yesterday

lbdreyer:
> > The PY=36 tests failed with...
> 
> Because [SciTools/iris#4163](https://github.com/SciTools/iris/pull/4163) was merged yesterday

Thank you @trexfeathers !

TomDufall:
I can't use the GitHub suggestion feature as it's not a file edited in this PR, but don't forget to update the [project classifiers in setup.py](https://github.com/SciTools/iris-grib/blob/99d43447444e46f1a909d665ea7b4a89fb5c949d/setup.py#L85) for supported Python versions.

lbdreyer:
> I can't use the GitHub suggestion feature as it's not a file edited in this PR, but don't forget to update the [project classifiers in setup.py](https://github.com/SciTools/iris-grib/blob/99d43447444e46f1a909d665ea7b4a89fb5c949d/setup.py#L85) for supported Python versions.

Thank you @TomDufall ! I knew had missed something!

lbdreyer:
The py37 tests keep timing out as it fails to resolve the environment. I am looking into adding lockfiles such as was done in iris, and iris-esmf-regrid (https://github.com/SciTools-incubator/iris-esmf-regrid/pull/68)

lbdreyer:
I have now added some of the lockfile infrastructure, using the implementation in Iris and iris-esmf-regrid as my guides.

I decided to mimic how we "bootstrapped" the use of lockfiles for the iris v3.0.x branch (see https://github.com/SciTools/iris/pull/4154). I will extend this infrastructure to update the lockfiles (via a github action) in a later PR. For this we may have to consider running the update against all active branches (i.e. master, v0.1.7.x, etc). Note, for iris, _only the lockfiles on the master branch updated_.

Another thing to be aware of: as we are testing against both the master branch of iris and the latest release, this leads to complexity with the lockfiles. Options:
* Drop testing against the master branch
* Have two sets of lockfiles, one for when we test against the master branch, and one for when we test against the latest release
In this PR I have just created lockfiles based of the latest release of iris (from conda_forge). It just so happens that this works for both since the requirements have not have diverged.

I have tried to do the minimum in this PR, so there will be some follow up work:
* Implement github action to update lockfiles
* doctests are broken. They haven't been running. Issue to address it: # 274
* coverage testing needs to be switched back on: #276

jamesp (approved):
Looks great, nice one @lbdreyer.  Some nice improvements here over the original implementation.

Approved, are you ready for this to be merged?

lbdreyer:
Thanks for reviewing @jamesp!

I'm happy for this to be merged. 
Once this is in I will cut the 0.17.1 release

jamesp:
> Another thing to be aware of: as we are testing against both the master branch of iris and the latest release, this leads to complexity with the lockfiles. Options:
> 
> Drop testing against the master branch
> Have two sets of lockfiles, one for when we test against the master branch, and one for when we test against the latest release
In this PR I have just created lockfiles based of the latest release of iris (from conda_forge). It just so happens that this works for both since the requirements have not have diverged.

What are the benefits of testing against master/branch-a/branch-b vs. only testing against a released version?  There is a lot of complexity added by doing so, we should be clear what we are trying to achieve by doing so.

An alternative may be to have an integration test in iris that runs on new PRs that confirms whether changes break the latest release of iris-grib.

lbdreyer:
> What are the benefits of testing against master/branch-a/branch-b vs. only testing against a released version? There is a lot of complexity added by doing so, we should be clear what we are trying to achieve by doing so.
>
>An alternative may be to have an integration test in iris that runs on new PRs that confirms whether changes break the latest release of iris-grib.

I believe the intention of adding testing against master was so we get a heads up of breaking changes, but this relies the changes to go into the iris-master branch and not a version branch (e.g. v3.0.x). I suppose we don't get much value from it, at least not worth the complexity introduced by supporting, and breaks things unintentionally (e.g. iris dropping py36 support meant we had to drop py36 support here just to get the ci working. We could handle this more sophisticatedly, but again: even more complexity!

I also don't think it's necessary to test against master for all new iris-grib PRs. Just seems like overkill!

I would be in support of finding an alternative solution. Will raise an issue to address this, and we can have further discussions there.

jamesp:
@lbdreyer thanks for raising an issue to track this, seems like the best way to keep in on our radar.

Issue #280, opened 2021-06-25
Rename master to main


lbdreyer:
See https://github.com/github/renaming

As we have already started making this change to other repos in the SciTools (see [iris discussion](https://github.com/SciTools/iris/discussions/4186), [nc-time-axis](https://github.com/SciTools/nc-time-axis/branches))

If there are no objections, I'd like to push ahead by the end of next week

lbdreyer:
@pp-mo has kindly renamed the branch, so this can be closed.

trexfeathers:
Note that [README.md](https://raw.githubusercontent.com/SciTools/iris-grib/99d43447444e46f1a909d665ea7b4a89fb5c949d/README.rst) has several badges that still point to `master`.

PR #281, opened 2021-07-05
Update README.rst

pp-mo:
Following @trexfeathers comment [here](https://github.com/SciTools/iris-grib/issues/280#issuecomment-873882751)

Various references to 'master' branch remaining in README

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -19,11 +19,11 @@ terms of its `GNU LGPLv3 license <COPYING.LESSER>`_.
 
 \(C) British Crown Copyright 2010 - 2021, Met Office
 
-.. |CirrusCI| image:: https://api.cirrus-ci.com/github/SciTools/iris-grib.svg?branch=master
+.. |CirrusCI| image:: https://api.cirrus-ci.com/github/SciTools/iris-grib.svg?branch=main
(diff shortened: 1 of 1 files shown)


trexfeathers (approved):

trexfeathers:
Is this failure related to the fact that there are references to Iris `master` in the Nox file?

pp-mo:
> Is this failure related to the fact that there are references to Iris master in the Nox file?

Maybe ...
Let's try https://github.com/SciTools/iris-grib/pull/282

pp-mo:
> > Is this failure related to the fact that there are references to Iris master in the Nox file?
> Maybe ...
> Let's try #282

None of that.  It's because we were testing with Python 3.6 by default, and Iris is no longer supporting that.

Unfortunately, it's not the only problem, as the Py3.7 tests show.
We were also using the routine 'iris.util._array_slice_ifempty`, which [we removed from Iris, but no longer needed due to changes in iris._lazy_data]().

Will address in #282 (which I'm going to rename)

pp-mo:
> Will address in #282 (which I'm going to rename)

Now in #283
I think this one should wait for that..

Issue #277, opened 2021-06-08
Testing against Iris release and Iris master branch


lbdreyer:
In https://github.com/SciTools/iris-grib/pull/231, we introduced testing against both a release of iris-grib and the latest iris master branch.

There has been some concern about the complexity introduced by handling both cases (see #272 for more discussion on this, e.g. [this comment](https://github.com/SciTools/iris-grib/pull/272#issuecomment-856635033)) 

We could consider alternative solutions to how we ensure iris and iris-grib don't diverge to much

bjlittle:
@pp-mo This was delivered by #415, which tests against the latest `iris` available from `conda-forge` and from the latest source from the `main` branch of `iris`.

Issue #251, opened 2021-03-01
Make iris-grib black compliant


jamesp:
`iris-grib` currently fails to pass `black --check`.

pp-mo:
Superceded #384

PR #289, opened 2022-02-23
Mergeback v0.17.x back on to main

lbdreyer:
We forgot to merge the `v.0.17.x` feature branch onto main.

Note this is a bit messy due to some changes that have been made to both branches.

**Changes to `main` after we cut the `v0.17.x` branch:**
a) 665df606de7b966886cd2b74ca9421f3207be9d0
b) 99d43447444e46f1a909d665ea7b4a89fb5c949d
c) e78018c87849c60cb259d7b2294977cd265cb77e

**Changes to `v0.17.x` branch since it was cut:**
a) 07fe08a85c6526a2a722c9b264da3acc1d76b2d7
b) 7d84c180286ef1950893f5f1cc109898c481e0ee
c) 758da19881aeb5eeec0892362efb0c8112d8ba56 - this commit is the main source of confusion as it contains (a) and (b) from the above list ("Changes to `main` after we cut the `v0.17.x` branch"). It also contains __some__ of the changes included in commit (c)
d) 3618908c0f804e15eacb5aab931b91981321b82e
e) 731be65f58ea066995f900ed3e68c68b747b10fe

The main source of conflict was to do with .cirrus.yml and noxfile.py as they had slightly diverged on each branch, but I have hopefully resolved the outstanding issues


lbdreyer:
The CI jobs that are running against iris-main are failing:
- py37 job - fails because Iris-main has dropped support for py37. I have removed this job
- py38 job - fails because various changes in Iris break changes in iris-grib. I have a PR to fix this (#288) but that will go in after this. For now we just comment out this job

.cirrus.yml
```diff
@@ -96,14 +96,11 @@ linux_task_template: &LINUX_TASK_TEMPLATE
...
   matrix:
-    - name: "${CIRRUS_OS} tests: iris=repo-main python=3.7"
-      env:
-        PY_VER: 3.7
-        IRIS_SOURCE: "source"
-    - name: "${CIRRUS_OS} tests: iris=repo-main python=3.8"
-      env:
-        PY_VER: 3.8
-        IRIS_SOURCE: "source"
+    # TODO: iris-grib does not currently work with iris-main
```
pp-mo, line 99: I like that you reference #251 in the comment-out section below on doctests (lines 118-128)
Could we do something similar here?
It may mean making an issue to note the problem -- I think #277 also refers

pp-mo (approved):
Sorry for delay, I finally got there !
Apart from the tiny suggestion, I think this is all just fine.
Just reply something on that + I'll merge.

PR #296, opened 2022-04-14
Transverse Mercator support negative scanning

trexfeathers:
Have been able to confirm that this behaviour handles a file as expected, thanks to @MoseleyS and @edwardccsteele.

### DIFF
--- a/iris_grib/_load_convert.py
+++ b/iris_grib/_load_convert.py
@@ -829,12 +829,6 @@ def grid_definition_template_12(section, metadata):
     cs = icoord_systems.TransverseMercator(lat, lon, easting, northing,
                                            scale, geog_cs)
 
-    # This has not been tested with -x scanning, so raise an error in that
-    #  case.
-    scan = scanning_mode(section['scanningMode'])
-    if scan.i_negative:
-        raise TranslationError('Unsupported -x scanning')
-
     # Deal with bug in ECMWF GRIB API (present at 1.12.1) where these
     # values are treated as unsigned, 4-byte integers.
     x1 = fixup_int32_from_uint32(section['X1'])
@@ -872,6 +866,7 @@ def grid_definition_template_12(section, metadata):
 
     # Determine the lat/lon dimensions.
     y_dim, x_dim = 0, 1
+    scan = scanning_mode(section['scanningMode'])
     if scan.j_consecutive:
         y_dim, x_dim = 1, 0
 
--- a/iris_grib/_load_convert.py
+++ b/iris_grib/_load_convert.py
@@ -852,6 +852,22 @@ def grid_definition_template_12(section, metadata):
     check_range(x1, x2, section['Ni'], section['Di'])
     check_range(y1, y2, section['Nj'], section['Dj'])
 
+    # Further over-specification - the sequence of X1 & X2 is enough to
+    #  generate the sequence in the correct direction (also Y1 & Y2). All
+    #  scanningMode can do is add confusion; warn if there is inconsistency.
+    def validate_scanning(axis: str, stated: bool, encoded: bool):
+        def scan_str(scanning_bool):
+            return "positive" if scanning_bool else "negative"
+        if stated != encoded:
+            message = (
+                f"File states {scan_str(stated)} {axis} scanning, but encodes "
+                f"{scan_str(encoded)} {axis} scanning."
+            )
+            warnings.warn(message)
+    scan = scanning_mode(section['scanningMode'])
+    validate_scanning("X", not scan.i_negative, x1 < x2)
+    validate_scanning("Y", scan.j_positive, y1 < y2)
+
     x_points = np.linspace(x1 * CM_TO_M, x2 * CM_TO_M, section['Ni'])
     y_points = np.linspace(y1 * CM_TO_M, y2 * CM_TO_M, section['Nj'])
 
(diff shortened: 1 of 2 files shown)


trexfeathers:
Have expanded the scope to support negative scanning in either direction.

iris_grib/_load_convert.py
```diff
@@ -843,24 +843,18 @@ def grid_definition_template_12(section, metadata):
     # Check whether Di and Dj are as consistent as possible with that
     # interpretation - i.e. they are within 1cm.
     def check_range(v1, v2, n, d):
-        min_last = v1 + (n - 1) * (d - 1)
-        max_last = v1 + (n - 1) * (d + 1)
-        if not (min_last < v2 < max_last):
+        small = min(v1, v2)
+        large = max(v1, v2)
```
lbdreyer, line 847: The only thing that is slightly bothering me is if someone created a file where, for example, 
x1=0, x2=10 but they set the scanning mode as -x
but I suppose that isn't the job of this function to catch anyway. In the above case we would end up creating a coord of (0->10) and ignore the scanning mode.

Although it does point out another over-specification of the spec: why do you need to know the scanning direction if you have the location of the first and last point?

MoseleyS: My concern with this idea is that I think the data in question do not have the scanning mode set to "negative", so would fail this check.  I would prefer that these checks occurred after the user-supplied `callback` function, so that it becomes possible to fix known meta-data errors, but this probably isn't practical. As an alternative, perhaps a scanning-mode mismatch could be a Warning instead of an Exception?

pp-mo: FWIW my opinion: I think we already know that (a) the spec has redundancy and (b) files not fully conforming can be expected.  So it makes sense to do our best to make sense of what there is, but also emit a warning if not all information is consistent.

The existing code derives direction from Y1 + Y2, and spacing from "NY", and that works fine.  
It is not possible to have negative "DY" anyway, so Y1 and Y2 give a definitive scanning direction and you can't really get it wrong.  
I.E. if Y1 and Y2 were "intended" to be the other way round, this can only be indicated in the scanning direction.  But this would really violate the GDT description, which describes Y1 and Y2 as the "first" and "last" gridpoints.

So, I think we can safely just use Y1, Y2 and NY to define the coordinate.
I would then merely check that DY and scanning mode match the other evidence, and emit warnings if not.

pp-mo: (all of which goes equally for the X grid, obviously)

trexfeathers: febc77d7b6691e5f806b214d737ccce378253330

iris_grib/_load_convert.py
```diff
@@ -843,24 +843,34 @@ def grid_definition_template_12(section, metadata):
...
+        if not (min_last < large < max_last):
             raise TranslationError('Inconsistent grid definition')
     check_range(x1, x2, section['Ni'], section['Di'])
     check_range(y1, y2, section['Nj'], section['Dj'])
 
+    # Further over-specification - the sequence of X1 & X2 is enough to
+    #  generate the sequence in the correct direction (also Y1 & Y2). All
+    #  scanningMode can do is add confusion; warn if there is inconsistency.
+    def validate_scanning(axis: str, stated: bool, encoded: bool):
+        def scan_str(scanning_bool):
```
pp-mo, line 859: I don't think this code is black compliant, due to missing blank lines.
I'm not sure if this is a criticism or not, do we require it ?

iris_grib/_load_convert.py
```diff
@@ -843,24 +843,34 @@ def grid_definition_template_12(section, metadata):
...
     def check_range(v1, v2, n, d):
-        min_last = v1 + (n - 1) * (d - 1)
-        max_last = v1 + (n - 1) * (d + 1)
-        if not (min_last < v2 < max_last):
+        small = min(v1, v2)
+        large = max(v1, v2)
+        min_last = small + (n - 1) * (d - 1)
+        max_last = small + (n - 1) * (d + 1)
+        if not (min_last < large < max_last):
             raise TranslationError('Inconsistent grid definition')
```
pp-mo, line 851: I think there's a good opportunity here to make the message more explanatory.
It could produce messages like "file X grid is inconsistent : abs(X2 - X1) / Dj not within Nj +/- 1."

iris_grib/_load_convert.py
```diff
@@ -843,24 +843,34 @@ def grid_definition_template_12(section, metadata):
...
 
+    # Further over-specification - the sequence of X1 & X2 is enough to
+    #  generate the sequence in the correct direction (also Y1 & Y2). All
+    #  scanningMode can do is add confusion; warn if there is inconsistency.
+    def validate_scanning(axis: str, stated: bool, encoded: bool):
+        def scan_str(scanning_bool):
+            return "positive" if scanning_bool else "negative"
+        if stated != encoded:
+            message = (
+                f"File states {scan_str(stated)} {axis} scanning, but encodes "
```
pp-mo, line 863: This message could also be clearer :  IMHO it isn't clear what 'states' and 'encoded' mean.
Suggest messages more like 
"Gridpoint direction doesn't match X scanning mode : 'X2' > 'X1', but 'i_negative'=True."

I think/hope that can be generalised fairly easily?

pp-mo (changes requested):
I think the warning/error messages need to be clearer.
Otherwise OK now, I think.

iris_grib/_load_convert.py
trexfeathers: c2e482698ed5f115121db8530c8a7e1bb1d90848

iris_grib/_load_convert.py
trexfeathers: c2e482698ed5f115121db8530c8a7e1bb1d90848

pp-mo (approved):

pp-mo:
Thanks @trexfeathers  !

PR #318, opened 2023-01-22
Add loading+saving for non-supported fixed surfaces

lbdreyer:


iris_grib/_save_rules.py
```diff
@@ -942,12 +959,15 @@ def set_fixed_surfaces(cube, grib, full3d_cube=None):
         gribapi.grib_set(grib, "scaleFactorOfFirstFixedSurface", 0)
         gribapi.grib_set(grib, "scaledValueOfFirstFixedSurface", output_v)
         gribapi.grib_set(grib, "typeOfSecondFixedSurface", 255)
-        gribapi.grib_set(grib, "scaleFactorOfSecondFixedSurface", 255)
+        gribapi.grib_set(grib, "scaleFactorOfSecondFixedSurface",
+                         GRIB_MISSING_LONG)
```
lbdreyer, line 963: I was getting failing tests with these, so this should be GRIB_MISSING_LONG rather:
```
>>> import gribapi
>>> grib = gribapi.grib_new_from_samples("GRIB2")
>>> gribapi.grib_get(grib, "scaleFactorOfSecondFixedSurface")
2147483647
>>> gribapi.GRIB_MISSING_LONG
2147483647
```

iris_grib/tests/unit/save_rules/test_set_fixed_surfaces.py
```diff
@@ -118,10 +120,89 @@ def test_height_point(self, mock_set):
...
+        msg = r"coordinates were found of fixed surface type: \[151, 108\]"
+        with self.assertRaisesRegex(ValueError, msg):
+            set_fixed_surfaces(cube, grib)
+
+    def test_unhandled_vertical_axis(self):
+        grib = None
+        cube = iris.cube.Cube([0])
+        cube.add_aux_coord(
+            iris.coords.AuxCoord([450], attributes={'positive': 'up'}))
+        msg = r"coordinate\(s\) \('unknown'\) are not recognised or handled."
```
pp-mo, line 202: I feel it might be worthwhile extending the scope of the message match here, just for a bit more clarity.
I.E. at least as far as to say  "vertical-axis coordinate(s) ...".

iris_grib/tests/unit/save_rules/test_set_fixed_surfaces.py
```diff
@@ -118,10 +120,89 @@ def test_height_point(self, mock_set):
...
+            grib, "typeOfSecondFixedSurface"), 255)
+        self.assertEqual(gribapi.grib_get_long(
+            grib, "scaledValueOfSecondFixedSurface"), GRIB_MISSING_LONG)
+        self.assertEqual(gribapi.grib_get_long(
+            grib, "scaleFactorOfSecondFixedSurface"), GRIB_MISSING_LONG)
+
+    def test_unknown_vertical_bounded(self):
+        cube = iris.cube.Cube([0])
+        cube.add_aux_coord(
+            iris.coords.AuxCoord([450], bounds=np.array([900.0, 0.0]),
```
pp-mo, line 157: An upper bound value of 0.0 feels a bit like it _could_ be a default or a coincidence, when we check for it  in the output.
Could you maybe make it a more unlikely value, just for a bit more certainty?

pp-mo (changes requested):
Ok I raised a couple of points, but admit I'm being quite fussy.
I think the code is good + testing coverage is excellent !

pp-mo:
Regarding the failing tests, it seems clear that the test `integration.load_convert.test_data_section.TestDRT3.test_grid_complex_spatial_differencing` is now failing because when you load the 'missing_values' file, it now gets an extra 'unknown' vertical coordinate, which looks like this :
```
>>> path = tests.get_data_path(
...     ("GRIB", "missing_values", "missing_values.grib2")
... )
>>> cube = iris.load_cube(path)
>>> print(cube)
x_wind / (m s-1)                    (latitude: 73; longitude: 144)
    Dimension coordinates:
        latitude                             x              -
        longitude                            -              x
    Scalar coordinates:
        forecast_period             0 hours
        forecast_reference_time     2013-05-21 00:00:00
        time                        2013-05-21 00:00:00
        unknown                     1829.0
    Attributes:
        GRIB_PARAM                  GRIB2:d000c002n002
>>> print(cube.coord('unknown'))
DimCoord :  unknown / (unknown)
    points: [1829.]
    shape: (1,)
    dtype: float64
    attributes:
        GRIB_fixed_surface_type  102
>>> 
```
Checking what this _used_ to do, it did not previously have the extra "unknown" coordinate.
So, I guess this **_is_** intentional, but also it is a breaking change.
Are we comfortable with this this @lbdreyer, or do we need to control this somehow ?

iris_grib/_load_convert.py
```diff
@@ -1630,63 +1630,61 @@ def vertical_coords(section, metadata):
...
-                    metadata['aux_coords_and_dims'].append((coord, None))
-            else:
-                point = unscale(scaledValueOfFirstFixedSurface,
-                                scaleFactorOfFirstFixedSurface)
-                coord = DimCoord(point,
-                                 standard_name=fixed_surface.standard_name,
-                                 long_name=fixed_surface.long_name,
-                                 units=fixed_surface.units)
+                                     units=fixed_surface.units)
+                if coord.name() == 'unknown':
```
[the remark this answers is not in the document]
lbdreyer: It was being used a proxy which I suppose is a bit risky. Instead I have now implemented your latter suggestion

iris_grib/_save_rules.py
```diff
@@ -907,15 +907,29 @@ def set_fixed_surfaces(cube, grib, full3d_cube=None):
...
+        fs_v_coords = [coord for coord in cube.coords() if
+                       'GRIB_fixed_surface_type' in coord.attributes]
+        if len(fs_v_coords) > 1:
+            fs_types = [c.attributes['GRIB_fixed_surface_type']
+                        for c in fs_v_coords]
+            raise ValueError("Multiple vertical-axis coordinates were found "
+                             f"of fixed surface type: {fs_types}")
+        elif len(fs_v_coords) == 1:
+            v_coord = fs_v_coords[0]
+            warnings.warn("The vertical-axis coordinate unit may not be "
```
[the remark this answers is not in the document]
lbdreyer: I can't quite remember my thinking for this. I think I was concerned with the situation of a user creating the attribute themself. But I'm not very attached to this so I have chosen to just remove it.

lbdreyer:
> So, I guess this is intentional, but also it is a breaking change.
Are we comfortable with this @lbdreyer, or do we need to control this somehow to preserve behaviour ?
It's a bit awkward : I guess it would need the equivalent of an Iris FUTURE flag. We don't really have a policy for how to do all that in iris-grib AFAIK.

I don't see this as a problem, and I don't agree that it is a breaking change. We are now supporting the loading of something that we previously were skipping. The old behaviour was just due to Iris-grib's limitations. This _extends_ iris-grib's capability rather the changing it.

From a user's perspective, it may have an impact on how their cubes get loaded (e.g. a GRIB file with multiple heights now may be merged into a 3D cube rather than a list of 2D slices) but I see this as us _correcting_ the behaviour.

iris_grib/tests/unit/load_convert/test_vertical_coords.py
```diff
@@ -62,22 +72,48 @@ def _check(self, value, msg):
...
+                   'scaledValueOfFirstFixedSurface': 18000,
+                   'scaleFactorOfFirstFixedSurface': 0,
+                   'typeOfSecondFixedSurface': 108,
+                   'scaledValueOfSecondFixedSurface': 0,
+                   'scaleFactorOfSecondFixedSurface': 0}
+        vertical_coords(section, metadata)
+        coord = DimCoord(9000.0, bounds=[18000, 0],
+                         attributes={'GRIB_fixed_surface_type': 108})
+        expected = deepcopy(self.metadata)
+        expected['aux_coords_and_dims'].append((coord, None))
```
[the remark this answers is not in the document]
pp-mo: :+1:

iris_grib/tests/unit/save_rules/test_set_fixed_surfaces.py
pp-mo: :+1:

pp-mo (approved):

pp-mo:
Thanks @lbdreyer I think this is fine now.

Issue #321, opened 2023-03-02
Incorrect translation of forecastTimeUnits for values other than "hours" on bounded forecast_period coordinates


s-boardman:
For data which should have a bounded `forecast_period` where the end of the period is greater than t+255 Grib changes the representation to have units of `3 hours` rather than `hours`. This is because the `P1` and `P2` values are stored as a single octet so would overflow if greater than 255. To work around this grib expects the reading software to multiply the `P1/2` values by the units to produce the _actual_ values. Therefore units of `3 hours` will produce the actual value of `P1=252` and `P2=258` for the relevant sections of the below grib header (output by `grib_dump`):

```
18        unitOfTimeRange = 10 [3 hours (grib1/4.table) ]
19        P1 = 84
20        P2 = 86
21        timeRangeIndicator = 2 [Product with a valid time ranging between reference time + P1 and reference time + P2 (grib1/local/ecmf/5.table , grib1/5.table) ]
```

When processing data from ECMWF that went beyond t+255 I noticed that cubes with bounded forecast periods were being produced with incorrect `forecast_period` coordinate, eg for the above grib record:

`forecast_period: 253.0 3 hours, bound=(252.0, 254.0) 3 hours`

It should be:

`forecast_period: 255.0 hours, bound=(252.0, 258.0) hours`

Debugging the `iris_grib._grib1_load_rules` module showed that if the generated `_forecastTimeUnit` of a grib message is anything other than `hours` then the `forecast_period` coordinate will be incorrect due to the the `add_bounded_time_coords` [function](https://github.com/SciTools/iris-grib/blob/main/iris_grib/_grib1_load_rules.py#L187) incorrectly calculating the period and setting the units.

When calculating the period to be used as the `forecast_period` the following code first converts the `t_bounds` (time_bounds) into hours and then tries to repeat the conversion when calculating the `period` from the `t_bounds` by using the original unit as shown below:

```python
t_bounds = grib.phenomenon_bounds('hours')
period = Unit('hours').convert(t_bounds[1] - t_bounds[0], grib._forecastTimeUnit)
```

When the `_forecastTimeUnit` is `hours` this conversion has a null effect. However, if this is 3 hours, for example, then the conversion is incorrect:

```python
>>> Unit('hours').convert(258 - 252, "3 hours")
2.0
```

Given that the conversion to hours has already taken place by the `phenomenon_bounds('hours') call, the period should be calculated as a simple subtraction of the lower bound from the upper:

```python
[...]

PR #322, opened 2023-03-02
Correct bounded forecast period calculation

s-boardman:
Addresses #321 by correcting the calculation of the `period` in `forecast_period` coordinates with bounds. Currently the code tries to repeat a conversion of the values for the bounds into hours which results in an incorrect value. In addition the units of the `forecast_period` coordinate are not set to `hours` even though the values are converted to hours.

I suspect the testing I've done could be expanded / improved so I'm particularly keen for feedback which could help improve the coverage.


lbdreyer (approved):
This is brilliant! Thanks for your contribution @s-boardman !

What you have done with the testing is sufficient so I'm happy to just merge this in.

PR #357, opened 2023-10-12
replace deprecated `eccodes-python` PyPI package with new `eccodes` in core requirements

valeriupredoi:
PyPI's [`eccodes-python`](https://pypi.org/project/eccodes-python/) has now been deprecated and we are instructed to use `eccodes` instead (currently at 1.6.1 so please pin if needed).

Closes https://github.com/SciTools/iris-grib/issues/356

:beers:

PR #361, opened 2023-11-15
Fix for eccodes>=2.32.1 and python-eccodes>=1.6.1

bjlittle:
This pull-request updates `iris-grib` to cater within changes in the latest version of `eccodes`.

Due to this change in `eccodes` behaviour, I've defensively introduces a minimum pin for both `eccodes` and `python-eccodes`.

Note that, this PR is targeting the `auto-update-lockfiles` branch, which is the base branch of pull-request #354.

### DIFF
--- a/iris_grib/_save_rules.py
+++ b/iris_grib/_save_rules.py
@@ -315,9 +315,9 @@ def latlon_points_irregular(cube, grib):
     # So, this only works at present if the x and y dimensions are **equal**.
     lon_values = x_coord.points / _DEFAULT_DEGREES_UNITS
     lat_values = y_coord.points / _DEFAULT_DEGREES_UNITS
-    eccodes.codes_set_array(grib, 'longitudes',
+    eccodes.codes_set_array(grib, 'longitude',
                             np.array(np.round(lon_values), dtype=np.int64))
-    eccodes.codes_set_array(grib, 'latitudes',
+    eccodes.codes_set_array(grib, 'latitude',
                             np.array(np.round(lat_values), dtype=np.int64))
 
 
--- a/iris_grib/message.py
+++ b/iris_grib/message.py
@@ -413,17 +413,13 @@ class Section:
             if key == 'numberOfSection':
                 value = self._number
             else:
-                keys = [key]
-                # Also check known alias alternatives
-                if key in KEY_ALIAS:
-                    keys.append(KEY_ALIAS[key])
-                keys = np.array(keys)
-                found = [k in self._keys for k in keys]
-                if not any(found):
-                    emsg = f"{key} not defined in section {self._number}"
-                    raise KeyError(emsg)
-                # Take the first valid key, original or alias.
-                key = keys[found][0]
+                if key not in self._keys:
+                    key2 = KEY_ALIAS.get(key)
+                    if key2 and key2 in self._keys:
+                        key = key2
+                    else:
+                        emsg = f"{key} not defined in section {self._number}"
+                        raise KeyError(emsg)
                 value = self._get_key_value(key)
 
             self._cache[key] = value
--- a/iris_grib/message.py
+++ b/iris_grib/message.py
@@ -20,6 +20,14 @@ from iris.exceptions import TranslationError
 
 _SUPPORTED_GRID_DEFINITIONS = (0, 1, 5, 10, 12, 20, 30, 40, 90, 140)
 
+# Alias names for eccodes spatial computed keys.
+KEY_ALIAS = {
+    'latitude': 'latitudes',
+    'longitude': 'longitudes',
+    'latitudes': 'latitude',
+    'longitudes': 'longitude',
+}
+
 
 class _OpenFileRef:
     """
@@ -404,12 +412,22 @@ class Section:
         if key not in self._cache:
             if key == 'numberOfSection':
                 value = self._number
-            elif key not in self._keys:
-                raise KeyError('{!r} not defined in section {}'.format(
-                    key, self._number))
             else:
+                keys = [key]
+                # Also check known alias alternatives
+                if key in KEY_ALIAS:
+                    keys.append(KEY_ALIAS[key])
+                keys = np.array(keys)
+                found = [k in self._keys for k in keys]
+                if not any(found):
+                    emsg = f"{key} not defined in section {self._number}"
+                    raise KeyError(emsg)
+                # Take the first valid key, original or alias.
+                key = keys[found][0]
                 value = self._get_key_value(key)
+
             self._cache[key] = value
+
         return self._cache[key]
 
     def __setitem__(self, key, value):
@@ -437,6 +455,7 @@ class Section:
                        'satelliteNumber', 'instrumentType',
                        'scaleFactorOfCentralWaveNumber',
                        'scaledValueOfCentralWaveNumber',
+                       'longitude', 'latitude',
                        'longitudes', 'latitudes')
         if key in vector_keys:
             res = eccodes.codes_get_array(self._message_id, key)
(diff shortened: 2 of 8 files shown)


iris_grib/_save_rules.py
```diff
@@ -315,9 +315,9 @@ def latlon_points_irregular(cube, grib):
     # So, this only works at present if the x and y dimensions are **equal**.
     lon_values = x_coord.points / _DEFAULT_DEGREES_UNITS
     lat_values = y_coord.points / _DEFAULT_DEGREES_UNITS
-    eccodes.codes_set_array(grib, 'longitudes',
+    eccodes.codes_set_array(grib, 'longitude',
                             np.array(np.round(lon_values), dtype=np.int64))
-    eccodes.codes_set_array(grib, 'latitudes',
+    eccodes.codes_set_array(grib, 'latitude',
```
bjlittle, line 320: I'm not able to easily pin-point the [change in eccodes](https://confluence.ecmwf.int/display/ECC/History+of+Changes) that is the source of this difference :disappointed:

iris_grib/message.py
```diff
@@ -20,6 +20,14 @@
 
 _SUPPORTED_GRID_DEFINITIONS = (0, 1, 5, 10, 12, 20, 30, 40, 90, 140)
 
+# Alias names for eccodes spatial computed keys.
+KEY_ALIAS = {
+    'latitude': 'latitudes',
+    'longitude': 'longitudes',
+    'latitudes': 'latitude',
+    'longitudes': 'longitude',
+}
```
bjlittle, line 29: Confusingly, it appears that both the singular and plural computed keys are valid, but on for certain grids ... well, that's what the testing is telling me, and that has driven this particular change.

iris_grib/message.py
```diff
@@ -404,12 +412,22 @@ def __getitem__(self, key):
...
+                # Also check known alias alternatives
+                if key in KEY_ALIAS:
+                    keys.append(KEY_ALIAS[key])
+                keys = np.array(keys)
+                found = [k in self._keys for k in keys]
+                if not any(found):
+                    emsg = f"{key} not defined in section {self._number}"
+                    raise KeyError(emsg)
+                # Take the first valid key, original or alias.
+                key = keys[found][0]
```
bjlittle, line 426: See this [comment](https://github.com/SciTools/iris-grib/pull/361/files#r1394257805)

requirements/core.txt
```diff
@@ -1,4 +1,4 @@
 # Core dependencies.
 
 scitools-iris>=3.0.2
-eccodes-python
+eccodes>=1.6.1
```
bjlittle, line 4: See #357 - I'm pulling this change into this PR even though it's already on `main`, just for consistancy.

iris_grib/message.py
```diff
@@ -437,6 +455,7 @@ def _get_key_value(self, key):
                        'satelliteNumber', 'instrumentType',
                        'scaleFactorOfCentralWaveNumber',
                        'scaledValueOfCentralWaveNumber',
+                       'longitude', 'latitude',
                        'longitudes', 'latitudes')
```
bjlittle, line 459: Note that, removing `'longitudes', 'latitudes'` results in the tests still working, but I'm nervous about doing this as the whole foundation of this PR is unclear.

pp-mo: Totally agree.
I think this is the safest approach

requirements/core.txt
pp-mo: I don't understand the need for this.
**_Why_** is it on main? 
AFAICT, we have changed the name there, but _not_ pinned the version.

And surely, the new code will anyway still work in the "old world" ??

requirements/py310.yml
```diff
@@ -11,7 +11,8 @@ dependencies:
 
 # Core dependencies.
   - iris>=3.0.2
-  - python-eccodes
+  - python-eccodes>=1.6.1
```
pp-mo, line 14: ? not needed ?
as above

requirements/py311.yml
```diff
@@ -11,7 +11,8 @@ dependencies:
 
 # Core dependencies.
   - iris>=3.0.2
-  - python-eccodes
+  - python-eccodes>=1.6.1
+  - eccodes>=2.32.1
```
pp-mo, line 15: ..as above

requirements/py39.yml
```diff
@@ -11,7 +11,8 @@ dependencies:
 
 # Core dependencies.
   - iris>=3.0.2
-  - python-eccodes
+  - python-eccodes>=1.6.1
+  - eccodes>=2.32.1
```
pp-mo, line 15: ..as above

iris_grib/message.py
pp-mo, line 426: I'm sure this works but, given that we only support a _single_ alias for any key, I'm finding this even-handed "there are a list of possibles" approach a bit obscure.
Can't we just "if then if else fail" it ?

```suggestion
                if key not in self._keys:
                    key2 = KEY_ALIAS.get(key)
                    if key2 and key2 in self._keys:
                        key = key2
                    else:
                        emsg = f"{key} not defined in section {self._number}"
                        raise KeyError(emsg)
```

Not promising this definitely works (!)

iris_grib/_save_rules.py
pp-mo: > "why?"

At a guess, they are trying to increase consistency between different templates.
But clearly, only in an incremental way.  
Which makes rather a mess -- and they seem to not really care about the difference between a behavioural enhancement and a "breaking" change.

I guess they _could_ have supported either key on anything, as your change implements
😪

pp-mo (changes requested):
Thanks for taking this on @bjlittle.

No problem with this approach, basically.
-- IMHO the solution only looks untidy because the _problem_ space is such a mess.

I pushed have back on some details + suggestions 
-- please see what you think

requirements/core.txt
bjlittle: Nope. The changes here don't work in the old world, hence the need for a minimum pin.

Plus, on PyPI, ECMWF decided to rename the package.

requirements/py310.yml
bjlittle: Yup, it is, see [comment](https://github.com/SciTools/iris-grib/pull/361#discussion_r1394813547) above.

requirements/py311.yml
bjlittle: ditto ... as above 😉

requirements/py39.yml
bjlittle: ditto ... as above 😉

bjlittle:
@pp-mo Back to you fella 👍

iris_grib/_save_rules.py
bjlittle: @pp-mo BTW this is the **breaking change** in the `eccodes` API that is forcing us to introduce a minimum pin.

i.e., `longitude`/`latitude` for `codes_set_array` works for `python-eccodes>=1.6.1,eccodes>=2.32.1`, otherwise it must be `longitudes`/`latitudes`.

requirements/core.txt
bjlittle: Additionally, see this [comment](https://github.com/SciTools/iris-grib/pull/361/files#r1395419311) regarding the breaking change in the `eccodes` API 😞

iris_grib/_save_rules.py
pp-mo: Understood.
We just suck it up -- no real choice !

bjlittle:
@pp-mo Awesome, thanks! 🍻

PR #362, opened 2023-11-16
Auto-generate zero data for missing data payload

bjlittle:
This pull-request extends the capability of the `iris_grib.message._DataProxy` to detect and auto-generate missing data payload for a GRIB2 message.

The scenario is that the message data payload is all zero, and rather than serialize this within the message, thus increasing the footprint, an application can generate the data at runtime whenever the Data Representation Section [5] has `bitsPerValue=0` and an empty Data Section [7].

Failure to support this extension results in no encoded Data Section [7] within a `iris_grib.message._RawGribMessage` causing a `KeyError` traceback within the user script. The impact of this is that an N-D Iris cube can be loaded but whenever at least one GRIB2 message has such missing data, touching the cube data will result in this failure.

Reference: https://github.com/SciTools/iris-test-data/pull/89

### DIFF
--- a/iris_grib/message.py
+++ b/iris_grib/message.py
@@ -237,14 +237,29 @@ class _DataProxy:
         return bitmap
 
     def __getitem__(self, keys):
-        # NB. Currently assumes that the validity of this interpretation
+        # N.B. Assumes that the validity of this interpretation
         # is checked before this proxy is created.
 
         message = self.recreate_raw()
         sections = message.sections
+        data = None
+
+        if 5 in sections:
+            # Data Representation Section.
+            if sections[5]["bitsPerValue"] == 0:
+                # Auto-generate zero data of the expected shape and dtype, as
+                # there is no data stored within the Data Section of this GRIB
+                # message. Also flatten the result to 1-D for potential bitmap
+                # post-processing.
+                data = np.ravel(np.zeros(self.shape, dtype=self.dtype))
+
+        if data is None:
+            # Data Section.
+            data = sections[7]["codedValues"]
+
+        # Bit-map Section.
         bitmap_section = sections[6]
         bitmap = self._bitmap(bitmap_section)
-        data = sections[7]['codedValues']
 
         if bitmap is not None:
             # Note that bitmap and data are both 1D arrays at this point.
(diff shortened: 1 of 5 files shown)


bjlittle:
@pp-mo As a result of this fix, we now have "unexpected successes" in the tests regarding pickling.

We could now simply not mark these tests are expected failures. 

How would you like to proceed?

bjlittle:
Out of curiosity, I remove the `@unittest.expectedFailure` decorators ...

iris_grib/message.py
```diff
@@ -237,14 +237,29 @@ def _bitmap(self, bitmap_section):
...
     def __getitem__(self, keys):
-        # NB. Currently assumes that the validity of this interpretation
+        # N.B. Assumes that the validity of this interpretation
         # is checked before this proxy is created.
 
         message = self.recreate_raw()
         sections = message.sections
+        data = None
+
+        if 5 in sections:
```
bjlittle, line 247: This defensive check is necessary, as we have tests with no encoded section 5 entries in the message.

Otherwise we generate a `KeyError` when attempting an unprotected `sections[5]` look-up.

pp-mo:
> We could now simply not mark these tests are expected failures.

After some chat, I think this is OK.
I'll do some back-fill to check what the error USED to be, but I suspect we just found that latest-eccodes has fixed it (= made the pickling possible)

iris_grib/tests/integration/load_convert/test_data_section.py
```diff
@@ -68,5 +69,16 @@ def test_grid_complex_spatial_differencing(self):
         self.assertCMLApproxData(cube)
 
 
+class TestDataProxy(tests.IrisGribTest):
+    def test_data_representation__no_bitsPerValue(self):
+        """Ensures that zero data is auto-generated."""
+        path = tests.get_data_path(
```
pp-mo, line 75: Can we just add a comment, that this test file contains an empty data section ?

pp-mo (approved):
Unless you want to follow a tiny suggestion.. 
good to go I think

iris_grib/tests/integration/load_convert/test_data_section.py
bjlittle: @pp-mo See https://github.com/SciTools/iris-grib/pull/362/commits/92435d865722e54382c7d530b189bc6a46d0b703

pp-mo: Wow, that's comprehensive !

bjlittle:
@pp-mo Sweet, cheers 💯

Issue #365, opened 2023-11-16
Documentation Failing to build on RTD


bjlittle:
The documentation is failing to build, [here](https://readthedocs.org/projects/iris-grib/builds/22578130/).

Previously, the documentation hasn't been built for over 20 months :cry:

tkknight:
This is disappointing, clearly this repo needs more ownership.

tkknight:
PR to fix the docs build: https://github.com/SciTools/iris-grib/pull/366

pp-mo:
Fixed in #https://github.com/SciTools/iris-grib/pull/370

PR #366, opened 2023-11-17
Fix the docs building on readthedocs

tkknight:
The docs have not been built successfully for a while.  This updates the `.readthedocs.yml` file.

Updates based upon the Iris config https://github.com/SciTools/iris/blob/main/.readthedocs.yml.  I think there was no need for the post checkout actions though so we removed.

Further tweaks may be needed but this will get it building cleanly.

I have also enabled readthedocs building the docs for PR's so this the docs should not go stale again.

Fix for https://github.com/SciTools/iris-grib/issues/365.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,2 +1,32 @@
 conda:
-    file: requirements/readthedocs.yml
+version: 2
+
+build:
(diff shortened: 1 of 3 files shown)


pp-mo (approved):
This looks great.

PR #371, opened 2023-12-22
Update contributing guidelines in line with v5 SciTools CLA

trexfeathers:
See SciTools/.github#19 for details.

PR #359, opened 2023-11-13
Relicense from LGPL-3 to BSD-3

lbdreyer:
All authors of the current codebase have signed the new CLA (SciTools CLA v4) which allows relicensing to other OSI approved licenses.

I have followed the pattern I used in https://github.com/SciTools/cf-units/pull/331.

Grepping for "LGPL", "GPL" "GNU" and "COPYING" should now return no results.

PR #373, opened 2024-01-12
Investigate unexpected licence check failures.

pp-mo:
I can't  any difference between the licence headers of `test_grid_definition_template_40.py` and `test_grid_definition_template_140.py`, but the latter is failing.

The header check only runs on files changed in the current commit (or something like).
So we shall see whether this fails on the old file (previously apparently OK) or not ...

### DIFF
--- a/iris_grib/tests/test_license_headers.py
+++ b/iris_grib/tests/test_license_headers.py
@@ -66,20 +66,29 @@ class TestLicenseHeaders(unittest.TestCase):
             or cannot be found by subprocess, an IOError may also be raised.
 
         """
-        # Check the ".git" folder exists at the repo dir.
-        if not os.path.isdir(os.path.join(REPO_DIR, '.git')):
-            raise ValueError('{} is not a git repository.'.format(REPO_DIR))
 
-        # Call "git whatchanged" to get the details of all the files and when
-        # they were last changed.
-        output = subprocess.check_output(['git', 'whatchanged',
-                                          "--pretty=TIME:%ct"],
-                                         cwd=REPO_DIR)
-        output = output.decode().split('\n')
-        res = {}
-        for fname, dt in TestLicenseHeaders.whatchanged_parse(output):
-            if fname not in res or dt > res[fname]:
-                res[fname] = dt
+        # HACKED!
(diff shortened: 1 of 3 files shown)


pp-mo:
Weirder and weirder...

In this PR, it **_still_** complains about the file "..._140.py".
But that file is not changed, so should **_not_** get tested here IIUC.

And still I can detect no difference in the header lines between this and the "..._40.py" one.

pp-mo:
The latest results do seem to show that there is **_not_** anything wrong in the "test_..._140.py" file,
since it does not appear alongside "test_..._20.py" in the list of failing ones shown in the test output.

So far, I am stumped by this.  There must be something wrong with how the test is working.
@trexfeathers  can you make any sense of what might be happening here ?

trexfeathers:
> The latest results do seem to show that there is **_not_** anything wrong in the "test_..._140.py" file, since it does not appear alongside "test_..._20.py" in the list of failing ones shown in the test output.
> 
> So far, I am stumped by this. There must be something wrong with how the test is working. @trexfeathers can you make any sense of what might be happening here ?

I think you're looking in the wrong place. The failure is in `test_grid_definition_template_140.py` within `save_rules/` (not `load_convert/`). The one in save rules does indeed have the wrong licence header:

https://github.com/SciTools/iris-grib/blob/831ea257d82c3f0d7ec6a840f57654de9979b0a6/iris_grib/tests/unit/save_rules/test_grid_definition_template_140.py#L1-L5

This got merged unnoticed because #343 was raised before #359, but was merged AFTER it. We have got used to the excellent functionality of GHA, which gives us CI against a simulated merge, whereas Cirrus runs CI on the branch itself. This meant that #343 passed, because the licence header test on #343 was still asserting for the old header.

I assumed that the failures on #371 couldn't possibly be real, since I had not modified the file in question. But the failures were real, they just hadn't been noticed before. I'm probably going to run out of time to raise a fix PR this evening, but I'd be happy to merge one if you raised it.

pp-mo:
Looking at the wrong file !
There are **two** files called `test_grid_definition_template_140.py`, 
I have been looking at `tests/unit/load_convert`, 
but the one that matter is in `tests/unit/save_rules`

PR #405, opened 2024-04-09
Save GDT20 - Polar Stereographic - take 2

trexfeathers:
Closes #122

I believe the outstanding requests have been addressed:

- [x] https://github.com/SciTools/iris-grib/pull/122#discussion_r195443839
  - ff8fc0d5b496582d404f35a8ac597363f0b00970
  - 8f735bfd38397a3b97de09809b34add87c39063f
  - 756bb56495b39f4b08f46abae11333c0c8b34dcb
- [x] https://github.com/SciTools/iris-grib/pull/122#issuecomment-397318338
  - 26eb55ad4922dd6451794d5767a7cee58c1f24ae
- [x] https://github.com/SciTools/iris-grib/pull/122#issuecomment-470574776
  - 36dd8efba695a3060b150323d737531aa5644430
  - 6b0d2a1bc69645e370141e855ee1a7cf20957d29


trexfeathers:
@pp-mo I'm looking at this comment - https://github.com/SciTools/iris-grib/pull/122#issuecomment-397318338:

>I'm thinking that at least one of NorthPolar/SouthPolar will currently fail a save-then-load roundtrip check.
>That's a good reason for adding one before merging this.

Could you offer some guidance on this? There is a new integration test but I suspect that's not what you meant.

pp-mo:
> @pp-mo I'm looking at this comment - [#122 (comment)](https://github.com/SciTools/iris-grib/pull/122#issuecomment-397318338):
> 
> > I'm thinking that at least one of NorthPolar/SouthPolar will currently fail a save-then-load roundtrip check.
> > That's a good reason for adding one before merging this.
> 
> Could you offer some guidance on this? There is a new integration test but I suspect that's not what you meant.

I think what this meant is that we should also test with **_Iris_** `iris.coord_systems.PolarStereographic` coordinates systems -- i.e. as well as `iris.coord_systems.Stereographic`.

In CF terms, a "polar_stereographic" grid mapping must have latitude_of_projection_origin  = +90 or  -90.  It's effectively just a special case of "stereographic", but I guess the polar case was put in first so they needed to keep it.
Iris completely follows CF in this, in having a separate PolarStereographic coordinate system type, presumably so that we can sensibly distinguish "polar" ones, but the PolarStereographic class actually [_inherits_  from the Stereographic class](https://github.com/SciTools/iris/blob/e26bfaf93dd06f82f195ee303960dc3ae03b1d09/lib/iris/coord_systems.py#L1073-L1125), and only adds a distinct type and a modified constructor -- no different behaviour.

So, in terms of iris-grib, I think my old comment meant that data with an Iris PolarStereographic coordinate system (either north or south) _should_ be handled by iris-grib, being treated as a generic 'Stereographic' case.  But that "would have failed", at some point, for some reason.  

So it's probably still worth doing a roundtrip cube-grib-cube test on both North and South cases, and in each case when presented as either a PolarStereographic or Stereographic Iris coordinate system.
Though obviously when loading back, you will always get a PolarStereographic Iris coordinate system (since AFAICT GRIB does not seem to provide a more general case).

pp-mo:
> when loading back, you will always get a PolarStereographic Iris coordinate system

Actually no, to my surprise.
we do it [like this](https://github.com/SciTools/iris-grib/blob/70cafb0054b75832984a5ba570bc1a7dee0bcc9f/iris_grib/_load_convert.py#L918-L927)

iris_grib/_save_rules.py
```diff
@@ -634,17 +632,61 @@ def grid_definition_template_30(cube, grib):
                       int(np.round(first_x / _DEFAULT_DEGREES_UNITS)))
     eccodes.codes_set(grib, "LaD", cs.central_lat / _DEFAULT_DEGREES_UNITS)
```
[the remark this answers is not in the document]
trexfeathers: Having discussed with @pp-mo, the current code is perfectly correct since this value is always identical to central latitude in the polar stereographic case. What _does_ need improving is preventing users from saving any other value, since they could conceivably create a `Stereographic` instance with an alternative `true_scale_lat` but which would never be saved - need a `TranslationError`.

iris_grib/_save_rules.py
```diff
@@ -604,22 +600,24 @@ def grid_definition_template_30(cube, grib):
...
+        msg = 'Irregular coordinates not supported for {!r}.'
+        raise TranslationError(msg.format(cs_name))
     eccodes.codes_set(grib, 'Dx', abs(x_step))
     eccodes.codes_set(grib, 'Dy', abs(y_step))
 
     horizontal_grid_common(cube, grib, xy=True)
 
-    # Transform first point into geographic CS
+    eccodes.codes_set(grib, 'resolutionAndComponentFlags',
+                      0x1 << _RESOLUTION_AND_COMPONENTS_GRID_WINDS_BIT)
```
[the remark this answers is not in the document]
trexfeathers: Misread this

iris_grib/_save_rules.py
```diff
@@ -604,22 +600,24 @@ def grid_definition_template_30(cube, grib):
...
     try:
         x_step = step(x_mm, atol=1)
         y_step = step(y_mm, atol=1)
     except ValueError:
-        msg = ('Irregular coordinates not supported for Lambert '
-               'Conformal.')
-        raise TranslationError(msg)
+        cs_name = cs.grid_mapping_name.replace('_', ' ').title()
+        msg = 'Irregular coordinates not supported for {!r}.'
+        raise TranslationError(msg.format(cs_name))
```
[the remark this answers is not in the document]
trexfeathers: 079fcfce4eb10c32c4b5e1cbd61227400e792e7c

iris_grib/_save_rules.py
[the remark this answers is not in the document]
trexfeathers: 079fcfce4eb10c32c4b5e1cbd61227400e792e7c

iris_grib/_save_rules.py
```diff
@@ -670,20 +683,81 @@ def grid_definition_template_30(cube, grib):
...
+    """
+    eccodes.codes_set(grib, "gridDefinitionTemplateNumber", 20)
+
+    # Retrieve the cube coord system.
+    cs = cube.coord(dimensions=[0]).coord_system
+
+    _perspective_projection_common(cube, grib)
+
+    # Is this a north or south polar stereographic projection?
+    if cs.central_lat == -90.0:
```
pp-mo, line 704: I think for safety I would allow a floating-point tolerance here.
Something like `np.allclose(cs.central_lat, -90.0)`, and similar
(possibly there are neater ways of saying it ?)

iris_grib/_save_rules.py
```diff
@@ -670,20 +683,81 @@ def grid_definition_template_30(cube, grib):
...
+        centre_flag = 0x80
+    elif cs.central_lat == 90.0:
+        centre_flag = 0x0
+    else:
+        emsg = (
+            "Bipolar and symmetric polar stereo projections "
```
pp-mo, line 710: Where do you get the class descriptions for this ?
I think it would be better to describe a case with central_lat not +/-90 just as "non-polar stereographic projections".  But perhaps there are other subtleties here that I'm not aware of ?

iris_grib/_save_rules.py
```diff
@@ -670,20 +683,81 @@ def grid_definition_template_30(cube, grib):
...
+        )
+        raise TranslationError(message)
+
+    if cs.scale_factor_at_projection_origin is not None:
+        message = (
+            f"{cs.scale_factor_at_projection_origin=} . iris_grib cannot "
+            "write scale_factor_at_projection_origin to a GRIB Template 3.20 "
+            "file."
+        )
+        raise TranslationError(message)
```
pp-mo, line 728: I think these 2 are basically the same case (which can be described in different ways),
and so would be better if you issued a common error message, 
e.g. "cannot save a Stereographic projection with a non-standard scale-factor".

In that case, I think you can just show all 3 of the parameters, to demonstrate why we are objecting.

iris_grib/_save_rules.py
```diff
@@ -670,20 +683,81 @@ def grid_definition_template_30(cube, grib):
...
+        raise TranslationError(emsg)
+
+    if cs.true_scale_lat and cs.true_scale_lat != cs.central_lat:
+        message = (
+            f"{cs.true_scale_lat=}, {cs.central_lat=} . iris_grib can "
+            "only write a GRIB Template 3.20 file where these are identical."
+        )
+        raise TranslationError(message)
+
+    if cs.scale_factor_at_projection_origin is not None:
```
pp-mo, line 722: I think this should also allow `all_close(cs.scale_factor_at_projection_origin, 1.0)`

pp-mo (changes requested):
A couple of suggestions, but looks very comprehensive to me.  Practically good to go.

iris_grib/_save_rules.py
trexfeathers: [ECMWF | Section 3 - Template 20 : Polar stereographic projection](https://codes.ecmwf.int/grib/format/grib2/templates/3/20/)

>| Octets | Key | Type | Content |
>| --- | --- | --- | --- |
>| ... | ... | ... | ... |
>| 64 | projectionCentreFlag | codeflag | Projection centre flag (See Flag Table 3.5) |
>| ... | ... | ... | ... |
>
>Notes:
>...
>( 4) Bit 2 of the projection flag is not applicable to the polar stereographic projection.

[ECMWF | Section 3 - Flag Table 5 : Projection Centre](https://codes.ecmwf.int/grib/format/grib2/ftables/3/5/)

>| Bit | Value | Meaning |
>| --- | --- | --- |
>| ... | ... | ... |
>| 2 | 1 | Projection is bi-polar and symmetric |

iris_grib/_save_rules.py
trexfeathers: abb29ad5963fc4c37d0d936077a973d11fa84a5b

iris_grib/_save_rules.py
trexfeathers: It's good information to know that these both constitute a similar problem - I didn't know that. But I worry that the user might also not know it, and so it is valuable to have the extra information about what specifically is wrong. So I have kept my original messages but added a common bit of information about these both being non-standard scale factors.

98068f48505aa91d351571b97d1cfde5fafa10cc

iris_grib/_save_rules.py
pp-mo: What "only one projection centre is used" I really don't understand.  I guess a polar projection always has a single centre, but it might not be at either the North or South pole.  So then "LaD" would be something other than +/-90, which we are not supporting as we aren't sure if + how it is legal.
All this is just guesswork, really.

trexfeathers: f793bdd5711c681f937cf0844590c36c0f05b58f

pp-mo: Great, I'll happily  accept that.
Thanks for your patience !

pp-mo (approved):

trexfeathers:
Thanks @pp-mo 😊

PR #407, opened 2024-04-10
Configure for mypy

### LINKED ISSUE #386: iris-grib MyPy (type checking)
https://learn.scientific-python.org/development/guides/repo-review/?repo=scitools%2Firis-grib

stephenworsley:
Closes #386

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,6 +86,9 @@ skip = "_build,*.css,*.ipynb,*.js,*.html,*.svg,*.xml,.git,generated"
 
 [tool.mypy]
 ignore_missing_imports = true
+warn_return_any = true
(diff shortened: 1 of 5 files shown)


trexfeathers:
Casually introduce MyPy 😎

pyproject.toml
```diff
@@ -0,0 +1,8 @@
+[tool.mypy]
```
ESadek-MO, line 1: Somewhere here, show_error_codes is being caught as deprecated. However, since it's not being used, and it's a minor issues, I think we let this go.

ESadek-MO (approved):
Thanks @stephenworsley! Left a comment for future sake, but I think this looks great!

PR #413, opened 2024-04-15
add iris-sample-data test dependency

bjlittle:
This pull-request adds the `iris-sample-data` package dependency, as required by doctests.


ESadek-MO (approved):
Thanks @bjlittle, looks ideal!

Issue #340, opened 2023-07-05
Migrate from Cirrus CI to GitHub Actions


trexfeathers:
For alignment with the rest of SciTools

**both tests and doctests**, please!
```[tasklist]
### Tasks  (attempt : drop if tricky)
- [ ] https://github.com/SciTools/iris-grib/issues/277
- [ ] https://github.com/SciTools/iris-grib/issues/417
```

PR #415, opened 2024-04-15
Convert tests and doctests from Cirrus to GHA

ESadek-MO:
Step 1 in #340, will have a separate PR for linkchecks.
Pull request is replicated from #409, due to rebase complications.

### DIFF
--- a/.github/workflows/composite/cartopy-cache/action.yml
+++ b/.github/workflows/composite/cartopy-cache/action.yml
@@ -1,0 +1,41 @@
+name: "cartopy-cache"
+description: "create and cache cartopy assets"
+
+#
+# Assumes the environment contains the following variables:
+#   - CONDA
+#
+inputs:
+  cache_build:
+    description: "conda environment cache build number"
+    required: false
+    default: "0"
+  cache_period:
+    description: "conda environment cache timestamp"
+    required: true
+  env_name:
+    description: "environment name"
+    required: true
+
+runs:
+  using: "composite"
+  steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/composite/cartopy-cache/action.yml
```diff
@@ -0,0 +1,41 @@
...
+        CARTOPY_FEATURE: https://raw.githubusercontent.com/SciTools/cartopy/v0.20.0/tools/cartopy_feature_download.py
+      shell: bash
+      run: |
+        # Require to explicitly activate the environment within the composite action.
+        source ${{ env.CONDA }}/etc/profile.d/conda.sh >/dev/null 2>&1
+        conda activate ${{ inputs.env_name }}
+        wget --quiet ${CARTOPY_FEATURE}
+        mkdir -p ${CARTOPY_SHARE_DIR}
+        # Requires a pre-installed version of cartopy within the environment.
+        python cartopy_feature_download.py physical --output ${CARTOPY_SHARE_DIR} --no-warn
```
bjlittle, line 41: @ESadek-MO This has changed for the better with `cartopy` v0.23+, see https://github.com/SciTools/cartopy/pull/2263

However, we can change this at a later date. The current implementation will still always work 👍

bjlittle (approved):
@ESadek-MO Let's bank this as-is.

We still need to inject the testing code coverage and the linkcheck, but these are additive changes.

Keen to see this banked so we can start getting mileage on it.

Great work 🍻

PR #419, opened 2024-04-16
Remove PEP8 test

trexfeathers:
PEP8 is outdated and doesn't understand modern principles such as typing, and is therefore getting in the way of current development. #384 is on its way which will replace anyway.


requirements/locks/py310-linux-64.lock
```diff
@@ -99,7 +99,6 @@ https://conda.anaconda.org/conda-forge/noarch/munkres-1.1.4-pyh9f0ad1d_0.tar.bz2
 https://conda.anaconda.org/conda-forge/noarch/nose-1.3.7-py_1006.tar.bz2#382019d5f8e9362ef6f60a8d4e7bce8f
 https://conda.anaconda.org/conda-forge/linux-64/openjpeg-2.5.2-h488ebb8_0.conda#7f2e286780f072ed750df46dc2631138
 https://conda.anaconda.org/conda-forge/noarch/packaging-24.0-pyhd8ed1ab_0.conda#248f521b64ce055e7feae3105e7abeb8
-https://conda.anaconda.org/conda-forge/noarch/pep8-1.7.1-py_0.tar.bz2#d0b5bc3aebbffcb175bf8a1d419acb79
```
bjlittle, line 102: @trexfeathers Did you just remove this line or rebuilt the locks from the new YAML?

I would have expected the `input_hash` to have also changed.

trexfeathers: Oh, thanks, I hadn't considered the `input_hash`. I did just remove the line because I figured any dependencies would be removed during the next lockfile update. In that spirit I'm inclined to just back out my lockfile changes.

trexfeathers: 545eaf412c1f8594ea624d2945e191345d1a4ae6

PR #420, opened 2024-04-16
Migrate to pytest

bjlittle:
This pull-request migrates away from the home-brew `nose` runner implementation to `pytest`.

Note that, the tests may be now run from the command line using `nox` or `pytest`. However, when using `pytest` from the CLI the `iris` environment variable `OVERRIDE_TEST_DATA_REPOSITORY` requires to be also configured.

I've not used the `pytest-xdist` plugin, as I didn't deem it in scope ATM; the existing runtime of the tests is relatively short.

In the future, we should look to use [filterwarnings](https://docs.pytest.org/en/stable/reference/reference.html#confval-filterwarnings) to control and manage the expected warnings raised during a test run.


pp-mo (approved):
Totally good with this !

Just one question : are the lockfiles re-resolved or manually edited ?
I can see they lose nose + gain a few new things, but no version change to existing packages.  I guess that must be "ok", but possibly not optimal ?

bjlittle:
@pp-mo The lock files were resolved using `conda-lock`, and result in `nose` being dropped and `pytest` being added.

We've recently updated the lock files in #419 (to remove `pep8`) so I wouldn't expect there to be much change in the other packages.

stephenworsley (approved):
Looks good!

bjlittle:
@pp-mo and @stephenworsley 

Awesome, thanks guys! 🥳

PR #424, opened 2024-04-16
add pytest-mock dependency

bjlittle:
This pull-request removes the `mock` package, which is not needed as it is now bundled with core python through `unittest`. 

It also adds the `pytest-mock` plugin package in preparation to support the eventual migration of our tests to `pytest`, given #420 has now landed.


stephenworsley (approved):
Looks good!

PR #425, opened 2024-04-16
fix ci-tests nose workaround

bjlittle:
This PR undoes a temporary workaround introduced by #415 for a limitation encountered when using the `nose` package on a GHA runner.

As we have now migrated away from `nose` and are using `pytest`, this workaround is now no longer necessary 👍


bjlittle:
👀 Ping @ESadek-MO 

I've undone this workaround before we forget 👍

stephenworsley (approved):
Looks good!

PR #436, opened 2024-04-18
add toml to manifest

bjlittle:
This pull-request adds `*.toml` files to the sdist manifest.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -12,3 +12,4 @@ exclude .pre-commit-config.yaml
 exclude .readthedocs.yml
 include *.md
 include *.py
+include *.toml


HGWright (approved):
nice spot to fix the failures.

trexfeathers:
@bjlittle perhaps you can educate me: I would have expected to `exclude` `.toml` files, since they won't be required as part of a user install?

I note that [Iris excludes `.ruff.toml` specifically](https://github.com/SciTools/iris/blob/6cc7c3df60f00689f5d4d5b3ed8ad218b90a59e0/MANIFEST.in#L17), and makes no mention at all of `pyproject.toml`, suggesting that maybe some special rules apply there.

---

Note that following two similar PR's, we now have this:

https://github.com/SciTools/iris-grib/blob/71703ae353abe26c224a6f3c7c46dac839ba896c/MANIFEST.in#L13-L16

I knew I should have checked `main` before merging #435!

PR #437, opened 2024-04-18
pep517: remove wheel dependency from build-system

bjlittle:
This pull-request removes the redundant inclusion of the `wheel` package as part of the `build-system`.

This is automatically included by the [PEP517](https://peps.python.org/pep-0517/) compliant backend.

Resolves the following `repo-review` non-compliance:

image


pp-mo (approved):
:+1:

PR #438, opened 2024-04-18
fix blacken-docs hook url

bjlittle:
This pull-request changes to source URL for the `blacken-docs` pre-commit hook, as recommended by `repo-review`.

Resolves the following `repo-review` non-compliance:

image


pp-mo (approved):
:+1:

PR #441, opened 2024-04-18
backout manifest inclusion

bjlittle:


stephenworsley (approved):
There might be a case to revisit the structure of MANIFEST.in in the future, but for now it makes sense to keep it consistent with Iris. I'm happy merging this.

Issue #421, opened 2024-04-16
Shift all the package code from `iris_grib` to `src/iris_grib`.


pp-mo:
**NOTE** disruptive (!!)
Don't do in the sprint -- do shortly afterwards when all (most) PRs are closed.

pp-mo:
#450 closes

PR #444, opened 2024-04-18
Adding setuptools_scm to iris grib for automatic versioning

### LINKED ISSUE #418: Adopt automatic versioning
Definitely need for automated builds [#285](https://github.com/SciTools/iris-grib/issues/285)

HGWright:
setuptools_scm added to pyproject, init, readthedocs and requirements, and then regenerating the lockfiles as its been added to the dependencies.

Closes #418


pyproject.toml
```diff
@@ -4,6 +4,7 @@
 # Defined by PEP 518
 requires = [
     "setuptools>=64",
+    "setuptools_scm>=7.0",
```
pp-mo, line 7: In Iris we have here ` "setuptools_scm[toml]>=7.0",`

Not clear if that can help.
Just trying to understand the source of the carnage here !

pp-mo: Another outside possibility is the lack of a 'lib' or 'src' directory.
That is not modern standard practice, perhaps it upsets scm ??
We have [this task](https://github.com/SciTools/iris-grib/issues/421) to shift the top-level folder , but I was put that off until after the sprint, as it will presumably break outstanding PRs.

trexfeathers: I don't think the `[toml]` thing should be a problem either, since @HGWright is already following the official guidance: [pypa/setuptools_scm: the blessed package to manage your versions by scm tags (github.com)](https://github.com/pypa/setuptools_scm)

pyproject.toml
trexfeathers, line 1: This needs to go:

https://github.com/SciTools/iris-grib/blob/c69489ec2b417368d7d84ac948374c7e2086805f/pyproject.toml#L64

.readthedocs.yml
```diff
@@ -4,6 +4,19 @@ build:
   os: ubuntu-20.04
   tools:
     python: mambaforge-4.10
+  jobs:
+    post_checkout:
+      # The SciTools/iris repository is shallow i.e., has a .git/shallow,
```
trexfeathers, line 9: Copypasta

pyproject.toml
```diff
@@ -60,13 +61,17 @@ license-files = ["LICENSE"]
...
 readme = {file = "README.rst", content-type = "text/x-rst"}
-version = {attr = "iris_grib.__version__"}
 optional-dependencies.all = {file = "requirements/all.txt"}
 optional-dependencies.test = {file = "requirements/test.txt"}
 
 [tool.setuptools.packages.find]
 include = ["iris_grib*"]
 
+[tool.setuptools_scm]
+write_to = "iris_grib/_version.py"
```
trexfeathers, line 71: You might have noticed that this new file is polluting your `git status`. We need an addition to the `.gitignore` file, like [this one in Iris](https://github.com/SciTools/iris/blob/7a3248668d57c7e6010dd8a9c4ee8602124f1bf9/.gitignore#L3-L4).

trexfeathers (changes requested):
I've installed this locally and it works 👍

Couple of small changes though:

trexfeathers (approved):

PR #446, opened 2024-04-19
Handle missing keys for GDT10 Mercator

### LINKED ISSUE #188: Incomplete validity checking in GDT 3.10
The [translation routine for gdt 3.10 (mercator grids)](https://github.com/SciTools/iris-grib/blob/v0.15.0/iris_grib/_load_convert.py#L755) does not check all termplate parameters.
Specifically, it doesn't check the scanningMode, nor *any* of the following parameters at the end of the template :  "Orientation of the grid", "longitudinal direction grid length", "latitudinal direction grid length" 
(I haven't checked what the actual key names of those in ecCodes would be)

For comparison, the translation routine for gdt 3.12 (transverse mercator grids) [***does*** check the scanningMode parameter](https://github.com/SciTools/iris-grib/blob/v0.15.0/iris_grib/_load_convert.py#L854), and errors if contains unsupported settings.

trexfeathers:
Closes #188


iris_grib/_load_convert.py
```diff
@@ -776,10 +776,35 @@ def grid_definition_template_10(section, metadata):
     # intersects the Earth
     standard_parallel = section["LaD"] * _GRID_ACCURACY_IN_DEGREES
 
+    if "orientationOfTheGrid" in section and not np.isclose(
```
pp-mo, line 779: What's the need for "if ... in section" here -- have I misunderstood something ?
By my understanding, if gdt is 10 then this key _should_ always be present, though it may have missing/None value ?

pp-mo (approved):
Other than tiny query -- LGTM

iris_grib/_load_convert.py
```diff
@@ -776,10 +776,35 @@ def grid_definition_template_10(section, metadata):
...
     cs = icoord_systems.Mercator(standard_parallel=standard_parallel, ellipsoid=geog_cs)
 
     # Create the X and Y coordinates.
     x_coord, y_coord, scan = _calculate_proj_coords_from_grid_lengths(section, cs)
+    final_x_point = x_coord.points[-1]
+    final_y_point = y_coord.points[-1]
+    if not (
+        np.isclose(section["longitudeOfLastGridPoint"], final_x_point)
+        and np.isclose(section["latitudeOfLastGridPoint"], final_y_point)
+    ):
```
trexfeathers, line 798: Experiments so far have produced many cases where this block is true, which makes me worried I need an extra factor in here like scaling or units or something. That's why I changed the PR back to draft.

trexfeathers: Still unsure about this one, but if @pp-mo has no objections I guess I'll take it out of draft.

pp-mo: I guess your experiments might indicate that more tolerance is required.
Since it's only a double-check, we can afford to be tolerant.  
Maybe something like atol=0.5 rtol=0.001 ?
So 10. +/- 0.5, 10,000. +/- 10.

PR #450, opened 2024-04-22
Migrate iris_grib to src/iris_grib

bjlittle:
This pull-request migrates the root `iris_grib` to `src/iris_grib`.


pp-mo (approved):
All looks good

PR #452, opened 2024-04-23
ignore noisy pull-requests with .git-blame-ignore-revs

bjlittle:
We've had some very noisy pull-requests banked recently, so it would be wise to adopt the use of [.git-blame-ignore-revs](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) so that the author of a noisy pull-request isn't blamed for the past sins of others :wink:


pp-mo (approved):
Can't predict how much more use this will get in future, but the suggested commits are helpful, and maintenance cost seems minimal

PR #456, opened 2024-04-25
adopt repo-review pre-commit hook

bjlittle:
We have some exceptions to the rule (some of which we can address), but it's best that we adopt the `repo-review` hook now to ensure that we don't regress.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,6 +108,7 @@ ignore = [
 # See https://mypy.readthedocs.io/en/stable/config_file.html
 ignore_missing_imports = true
 warn_unused_configs = true
+warn_unreachable = true
 enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
 exclude = [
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -108,6 +108,7 @@ ignore = [
 # See https://mypy.readthedocs.io/en/stable/config_file.html
 ignore_missing_imports = true
 warn_unused_configs = true
+warn_unreachable = true
```
bjlittle, line 111: This is a change recommended by `repo-review` :wink:

pp-mo (approved):
Looks cool !
I assume the simple "sp-repo-review ... Passed" line in the pre-commit actions output is all the evidence I need that this is working as intended

pp-mo:
Oooh, I missed something ...
Is there a **_badge_** we could adopt to highlight/promote the use of repo-check ?

bjlittle:
That would be nice, but not that I know of

PR #461, opened 2024-04-29
update .gitignore

bjlittle:
This pull-request updates the `.gitignore` base on https://github.com/github/gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,33 +1,171 @@
-*.py[co]
-
-# setuptools-scm
+# VCS by setuptools-scm
 _version.py
-
-# Packages
-*.egg
-*.egg-info
-dist
-build
-eggs
-parts
-bin
-var
-sdist
-develop-eggs
-.installed.cfg
-
-# test
-.nox/
-
-# Installer logs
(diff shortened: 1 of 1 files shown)


pp-mo (approved):
I do think this is all good, and in comparison with the existing includes all that currently covers 👍 

It's also very close to [github/gitignore/Python.gitignore](https://github.com/github/gitignore/blob/main/Python.gitignore)
So the templating idea does seem a really good one.

Given the existence of a very usable template elsewhere = https://github.com/github/gitignore as suggested, 
do we want to 
  - (1) reference that directly in our templating actions, or 
  - (2) make an scitools-specific snapshot copy of that, and maybe ..
    - (2a) .. template *that* (standard scitools version) against the latest in github/gitignore ?

pp-mo:
Ok from offline conversation with @trexfeathers , the templating approach does have value, but we will consider that separately + elsewhere (in Scitools/.github) https://github.com/SciTools/.github/issues/25

pp-mo:
Merging even though the iris-source tests fail : #463 or similar will eventually fix

PR #479, opened 2024-06-04
fix codespell

bjlittle:
This pull-request targets PR #478 to fix the new issues raised by the latest version of `codespell`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,8 +96,8 @@ exclude_lines = [
 
 [tool.codespell]
 # See https://github.com/codespell-project/codespell/tree/master?tab=readme-ov-file#using-a-config-file
-ignore-words-list = "alpha-numeric,degreee,discontiguities,lazyness,meaned,nin"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -96,8 +96,8 @@ exclude_lines = [
 
 [tool.codespell]
 # See https://github.com/codespell-project/codespell/tree/master?tab=readme-ov-file#using-a-config-file
-ignore-words-list = "alpha-numeric,degreee,discontiguities,lazyness,meaned,nin"
-skip = "_build,*.css,*.ipynb,*.js,*.html,*.svg,*.xml,.git,generated"
```
trexfeathers, line 100: I was a little surprised at this one, but I suppose we're not in control of it since it's a standard template, so ignoring spelling errors makes sense 👍

trexfeathers (approved):

pyproject.toml
bjlittle: Yeah, that was my take on it ... we don't "own" the code of conduct, so I'd prefer not to change it. 

If you're happy with that, then let's roll with it, thanks 😃

Issue #496, opened 2024-06-26
Remove MyPy `redundant-expr` error


trexfeathers:
https://github.com/SciTools/iris-grib/blob/f26971d12f7c2d8ae129ecae4439227c168905c9/pyproject.toml#L112

See https://github.com/SciTools/iris/pull/5956#discussion_r1653600177 for more detail.

PR #497, opened 2024-06-26
Remove MyPy redundant-expr error

githubalexliu:
https://github.com/SciTools/iris-grib/issues/496

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -138,6 +138,8 @@ xfail_strict = "True"
 ignore = [
     # https://learn.scientific-python.org/development/guides/style/#MY102
     "MY102",  # MyPy show_error_codes deprecated
+    # https://learn.scientific-python.org/development/guides/style/#MY105
+    "MY105",  # MyPy redundant-expr deprecated
     # https://learn.scientific-python.org/development/guides/style/#PC170
     "PC170",  # Uses PyGrep hooks
     # https://learn.scientific-python.org/development/guides/style/#PC180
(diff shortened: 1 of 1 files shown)


pp-mo:
Thanks @githubalexliu this is useful.
But, from the error, I think this change is still incomplete :  
It seems that since we also [adopted checking against the "repo-review" recommendations](https://github.com/SciTools/iris-grib/pull/456), this now complains because repo-review recommends that this MyPy option is enabled (!!).

So, I **_think_** you need to add something to the configuration of repo-review, presumably in [this section of pyproject.toml](https://github.com/SciTools/iris-grib/blob/f26971d12f7c2d8ae129ecae4439227c168905c9/pyproject.toml#L137-L147).
Presumably, we just need to add "MY105" to the list of ignores here, along with the existing MY102.

As it's so simple, I won't bother you with a PP-to-PR :  can you perhaps make this change + see if it works ?

pp-mo:
Thanks @githubalexliu 
I can't explain why the 3.12 tests failed like that, since nothing relevant changed (including the lockfiles)
And on re-spin went ok ??

Anyway, all good now I think !

pp-mo (approved):

pp-mo:
Oh I hadn't seen, we are also blocked on CLA.
Can you sing up for this, @githubalexliu ?

pp-mo:
Thanks @githubalexliu !
Sorry for delay, spent whole of last week focussing on SciPy2024

PR #504, opened 2024-07-05
Eccodes 2v36

pp-mo:
Fixes for a couple of backwards-incompatibilities introduced in eccodes version 2.36

~N.B. this also tested OK against 2.35, for which [these lines](https://github.com/SciTools/iris-grib/blob/a4c52810b003cc231b4fa3c79f55a2e1b93444fa/src/iris_grib/message.py#L28-L29) are required.~
~However, it's not clear how old an eccodes this will support, since `indicatorOfUnitOfTimeRange` is only aliassed on loading --  **_not_** in saving, where setting it always+only uses `indicatorOfUnitForForecastTime` (the new form).~

From their release page, v2.33 is when `indicatorOfUnitForForecastTime` was introduced, which we now write on saves.
Tested against eccodes v2.33 and v2.32 : confirmed the new code works with 2.33 but **_not_** 2.32.
As stated in whatsnew + updated requirements yml files.

N.B. the old name `indicatorOfUnitOfTimeRange` ***is*** still used in grib1 loading,  because that still requires it 
(!yuck!)

### DIFF
--- a/src/iris_grib/_save_rules.py
+++ b/src/iris_grib/_save_rules.py
@@ -977,7 +977,7 @@ def set_forecast_time(cube, grib):
     else:
         _, _, fp, grib_time_code = _missing_forecast_period(cube)
 
-    eccodes.codes_set(grib, "indicatorOfUnitOfTimeRange", grib_time_code)
+    eccodes.codes_set(grib, "indicatorOfUnitForForecastTime", grib_time_code)
     eccodes.codes_set(grib, "forecastTime", fp)
 
 
--- a/src/iris_grib/_save_rules.py
+++ b/src/iris_grib/_save_rules.py
@@ -230,12 +230,14 @@ def dx_dy(x_coord, y_coord, grib):
 
 
 def scanning_mode_flags(x_coord, y_coord, grib):
-    eccodes.codes_set_long(
-        grib, "iScansPositively", int(x_coord.points[1] - x_coord.points[0] > 0)
-    )
-    eccodes.codes_set_long(
-        grib, "jScansPositively", int(y_coord.points[1] - y_coord.points[0] > 0)
-    )
+    x_positive = x_coord.points[1] - x_coord.points[0] > 0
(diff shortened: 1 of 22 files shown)


pp-mo:
NOTE: merging this should unblock  subsequent PRs.
 * #499 
 * #500 
 * #501 
 * #502

src/iris_grib/_save_rules.py
```diff
@@ -230,12 +230,14 @@ def dx_dy(x_coord, y_coord, grib):
...
-        grib, "jScansPositively", int(y_coord.points[1] - y_coord.points[0] > 0)
-    )
+    x_positive = x_coord.points[1] - x_coord.points[0] > 0
+    y_positive = y_coord.points[1] - y_coord.points[0] > 0
+    scanningMode = 0
+    if not x_positive:
+        scanningMode |= 0x80  # "bit 1" has negative sense : set=decreasing
+    if y_positive:
+        scanningMode |= 0x40  # "bit2" has positive sense : set=increasing
+    eccodes.codes_set_long(grib, "scanningMode", scanningMode)
```
trexfeathers, line 240: There are still some references to `...ScansNegatively` and `...ScansPositively` in the Iris-grib codebase; apparently limited to GRIB-1 related code. Is this a problem?

trexfeathers (commented):
Thanks @pp-mo!

I used this for reference during my review: [v2.36.0 release notes](https://confluence.ecmwf.int/display/ECC/History+of+Changes#HistoryofChanges-Version2.36.0(June2024))

src/iris_grib/_save_rules.py
pp-mo: No, I think it's OK to _read_ them, but they can no longer be written.
And probably nest to keep GRIB1 code as it was, as with the old "indicatorOfUnitOfTimeRange" keyname which is still the version needed for GRIB1.

trexfeathers (approved):

PR #564, opened 2024-11-13
Initial upgrade to "Code QA Sprint" standards

pp-mo:
Updated pre-commit config, to be compared with recommendations in ["Scitools Code Quality Bonanza" project](https://github.com/orgs/SciTools/projects/43)

PR #568, opened 2024-11-19
Code qa refresh

pp-mo:
Replaces #564 

Don't understand why this was needed, since the branch location is the same.
Seems like GitHub got confused somehow by my adding a commit + then removing it

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -51,6 +51,32 @@ repos:
         # Duplicates Ruff W291 but also works on non-Python files.
     -   id: trailing-whitespace
 
+-   repo: https://github.com/adamchainz/blacken-docs
+    rev: 1.19.1
+    hooks:
+    -   id: blacken-docs
+        types: [file, rst]
+
+-   repo: https://github.com/codespell-project/codespell
+    rev: "v2.3.0"
+    hooks:
+    -   id: codespell
+        types_or: [asciidoc, python, markdown, rst]
+        additional_dependencies: [tomli]
+
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: 'v1.13.0'
+    hooks:
+    -   id: mypy
+        exclude: 'noxfile\.py|docs/conf\.py'
+
+-   repo: https://github.com/numpy/numpydoc
+    rev: v1.8.0
+    hooks:
+      - id: numpydoc-validation
+        exclude: "src/iris_grib/tests/|noxfile.py"
+        types: [file, python]
+
 -   repo: https://github.com/astral-sh/ruff-pre-commit
     rev: "v0.7.3"
     hooks:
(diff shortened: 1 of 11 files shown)


ukmo-ccbunney (approved):
As per your other PR, this looks good.
Remember to raise an issue for the ignored numpydoc tests.

ukmo-ccbunney (changes requested):
Actually - jsut spotted one thing:

`ruff-pre-commit` hook is now version 0.7.4, you've got it as 0.7.3

Can you update?

pp-mo:
> `ruff-pre-commit` hook is now version 0.7.4, you've got it as 0.7.3

Just added in a[ re-ordering of the pre-commit hooks to match the current template order](https://github.com/SciTools/iris-grib/pull/568/commits/0e537c3d843d8435211c79ca0320ba1004c9d22a).

ukmo-ccbunney:
> > `ruff-pre-commit` hook is now version 0.7.4, you've got it as 0.7.3
> 
> Just added in a[ re-ordering of the pre-commit hooks to match the current template order](https://github.com/SciTools/iris-grib/pull/568/commits/0e537c3d843d8435211c79ca0320ba1004c9d22a).

You are still running on 0.7.3 for Ruff though.

pp-mo:
> You are still running on 0.7.3 for Ruff though.

Sorry, missed including that somehow!

ukmo-ccbunney:
I wonder if rather than setting `strict=false` in your `[tools.mypy]` section, it would be better to set `strict=true` and capture all the failing error codes?

For iris-grib, this would require the following ignores in the pyproject.toml:

```
strict = true 
disable_error_code =  [
    "arg-type",
    "assignment",
    "attr-defined",
    "index",
    "misc",
    "name-defined",
    "no-untyped-call",
    "no-untyped-def",
    "override",
    "type-arg",
    "union-attr",
    "var-annotated",
]
```

These can then all be added to a new issue documenting the failing mypy tests for iris-grib?

What do you think?

pp-mo:
> I wonder if rather than setting `strict=false` in your `[tools.mypy]` section, it would be better to set `strict=true` and capture all the failing error codes?
> ... What do you think?

Well, here's one way of looking at what that does ...
```
$ mypy . --no-color-output | grep -oE "\[.*\]$" errs.txt | sort | uniq -c
      2 [<type>, <type>] = ...")  [var-annotated]
      2 [<type>] = ...")  [var-annotated]
     13 [Any, Any] | list[Any] | None" has no attribute "append"  [union-attr]
      2 [Any, Any] | list[Any] | None" is not indexable  [index]
     13 [Any, Any]" of "dict[Any, Any] | list[Any] | None" has no attribute "append"  [union-attr]
      1 [Any, Any]", variable has type "tuple[Any]")  [assignment]
      9 [Any, None]"; expected "tuple[Any, int]"  [arg-type]
      1 [Any]")  [index]
      1 [Any]", variable has type "tuple[Any, ...]")  [assignment]
      1 [Any]", variable has type "tuple[Any, Any]")  [assignment]
      1 [GenericConcreteGRIBCode]" has no attribute "argnames"  [attr-defined]
      1 [[Any, Any, Any, Any, Any], Any]" has no attribute "assert_called_once_with"  [attr-defined]
      1 [[Any, Any, Any, Any, Any], Any]" has no attribute "assert_called_with"  [attr-defined]
      1 [[Any, Any, Any, Any], Any]" has no attribute "assert_called_once_with"  [attr-defined]
      1 [[Any, Any], Any]" has no attribute "assert_called_with"  [attr-defined]
      1 [[Any], Any]" has no attribute "assert_called_with"  [attr-defined]
      1 [[Any], Any]" has no attribute "call_count"  [attr-defined]
      3 [assignment]
    191 [attr-defined]
      2 [index]
     15 [misc]
     12 [name-defined]
   1421 [no-untyped-call]
    964 [no-untyped-def]
      3 [override]
      1 [str] | None" has no attribute "groups"  [union-attr]
      1 [str]", variable has type "bytes")  [assignment]
      1 [str]", variable has type "type[int]")  [assignment]
      5 [type-arg]
     15 [var-annotated]
```

pp-mo:
> I wonder if rather than setting `strict=false` in your `[tools.mypy]` section, it would be better to set `strict=true` and capture all the failing error codes?

See https://github.com/SciTools/iris-grib/pull/568/commits/cdb78f53619953e41587b2868aa8ac3a80ee1143

ukmo-ccbunney (approved):
OK. I think we're good to go.

pp-mo:
Brilliant, thanks @ukmo-ccbunney !!

To complete this work : 
 * [ ] Small follow-up coming to rationalise the toml content
 * some issues for future removal of "ignores" 
   * [ ] ruff 
   * [ ] mypy 
   * [ ] numpydocs

PR #576, opened 2024-11-29
Fixes for recent merge-with-fails

pp-mo:
Looks like I made a mistake merging #567 

But it also seems that the required automatic fixes which a pre-commit run produces are not included in the pre-commit update PR.  Not sure if that is intended + should work, or what.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@
 # See https://github.com/scientific-python/cookie#sp-repo-review
 
 ci:
-  autofix_prs: false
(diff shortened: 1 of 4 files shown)


pp-mo:
Note: now enabled auto-fix on PRs

That won't actually stop merging that without due care + attention, like on #567.
But in this case it would have incorporated the fixes automatically.
For reference, autofixing is effectively on in Iris, since it is the default according to [the config docs](https://pre-commit.ci/#configuration)

trexfeathers (approved):

PR #578, opened 2024-11-29
Keyfix eccodes 2v38

pp-mo:
This also fixes a numpy2 problem.

Once working, should unblock the lockfile update (see #550).

### DIFF
--- a/src/iris_grib/_load_convert.py
+++ b/src/iris_grib/_load_convert.py
@@ -209,7 +209,7 @@ _MDI = None
 # Non-standardised usage for negative forecast times.
 def _hindcast_fix(forecast_time):
     """Return a forecast time interpreted as a possibly negative value."""
-    uft = np.uint32(forecast_time)
+    uft = np.array(forecast_time).astype(np.uint32)
     HIGHBIT = 2**30
 
     # Workaround grib api's assumption that forecast time is positive.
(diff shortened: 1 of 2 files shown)


src/iris_grib/_load_convert.py
```diff
@@ -209,7 +209,7 @@ def _masker(item):
 # Non-standardised usage for negative forecast times.
 def _hindcast_fix(forecast_time):
     """Return a forecast time interpreted as a possibly negative value."""
-    uft = np.uint32(forecast_time)
+    uft = np.array(forecast_time).astype(np.uint32)
```
trexfeathers, line 212: I'm gonna merge this anyway, but is there a particular reason you didn't do:

```suggestion
    uft = np.array(forecast_time, dtype=np.uint32)
```

trexfeathers (approved):

pp-mo:
Whoops I should have made this a Draft.

Unpinning numpy 2 reveals that the "other" problem **_isn't_** really fixed 
as it breaks at least the unit-test in ` src/iris_grib/tests/unit/load_convert/test__hindcast_fix.py::TestHindcastFix`

I am working on this, will sort it out + unblock the lockfile upate.

src/iris_grib/_load_convert.py
pp-mo: Recent versions of numpy (<2) were warning that the silent wrapping of values too big or small for the stated dtype was effectively deprecated + going to raise an error.
Which does happen with numpy2.

The 'astype' is a fix suggested by those warnings, which makes this code compatible with numpy 2.

**However** what went missing here is that this new code fails the tests when run under numpy 2, so further changes are still required to unpin our locks...

PR #590, opened 2025-01-02
Update to latest refresh-lockfiles shared action.

pp-mo:
Ironically, testing against latest-released Iris is now  causing us trouble also !

### DIFF
--- a/src/iris_grib/tests/unit/load_convert/test_product_definition_template_0.py
+++ b/src/iris_grib/tests/unit/load_convert/test_product_definition_template_0.py
@@ -14,6 +14,7 @@ import iris_grib.tests as tests
 
 from unittest import mock
 
+from iris import FUTURE as IRIS_FUTURE
 import iris.coords
 
 import iris_grib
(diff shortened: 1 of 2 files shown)


src/iris_grib/tests/unit/load_convert/test_product_definition_template_0.py
```diff
@@ -81,8 +82,14 @@ def test_generating_process_warnings(self):
         emit_warnings = convert_options.warn_on_unsupported
         try:
             convert_options.warn_on_unsupported = True
-            with mock.patch("warnings.warn") as warn:
-                product_definition_template_0(section_4(), metadata, rt_coord)
+            future_kwargs = {}
```
ESadek-MO, line 85: I assume you do it this way (via a dictionary) for a reason? I'm guessing it's a best practice that I'm not aware of.

ESadek-MO (approved):
I'm happy with this for the most part; just one question (based on my own ignorance).

src/iris_grib/tests/unit/load_convert/test_product_definition_template_0.py
pp-mo: Because it lets me always use a context block even when it's not doing anything, 
so that avoids repeating the code which appears inside the context, i.e. the alternative could be like ...
```
  if has_attr(IRIS_FUTURE, "date_microseconds"):
      with IRIS_FUTURE.context(date_microseconds=True):
          product_definition_template_0(section_4(), metadata, rt_coord)
  else:
      product_definition_template_0(section_4(), metadata, rt_coord)
  . . .
```
 -- that's really the only reason, cos I hate repeated code !

ESadek-MO: Oh, fair! Yep, I'm happy with that.

PR #597, opened 2025-01-07
Release notes for v0.21, to support latest ecCodes.

pp-mo:
As usual, we will create a new release branch just in case a bugfix might be needed.
This whatsnew update is the only change not already on main, and after release we'll subsequently merge this back. 

At that point, within the mergeback, we will also change the main documentation page version title to "v0.22 (unreleased)" to avoid confusion.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -3,7 +3,7 @@
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
 
-Iris-grib v0.20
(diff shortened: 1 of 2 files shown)


ESadek-MO (approved):
I'm happy with everything here!

I'm slightly cautious that I'm not used to a iris-grib release, but assuming you're happy you've done all the work there's no changes that I think are incorrect

Issue #202, opened 2020-02-24
Cannot pickle cubes with GRIB data


pp-mo:
#201 moves the grib-specific part of the iris 'test-pickle' from iris into iris-grib.
In principle, a cube should be picklable, but at present this doesn't work with grib data.
In fact a GribMessage itself cannot be successfully pickled + brought back.
So that needs fixing.

PR #620, opened 2025-03-06
Converted test_pickle_message to pytest, and removed reliance on Iris

ESadek-MO:
# Pull Request

Started off as a solution to recent test failures brought on from https://github.com/SciTools/iris/pull/6342, and ended up shuffling the furniture. 

This no longer relies on iris, which hopefully means iris-grib tests won't fail whenever something changes in iris. A couple of things to note:

- Should this still be placed in iris_integration tests? Presumably not.
- I've left in some commented code, which is bad form, but it's essentially a to-do. 
- The tests failing seem to be due to not running on pytest from what I can see, which begs the question, do we abandn this PR, and return to unit test and just fix the immediate failures?

### DIFF
--- a/src/iris_grib/tests/unit/message/test_pickle_message.py
+++ b/src/iris_grib/tests/unit/message/test_pickle_message.py
@@ -33,16 +33,16 @@ class TestPickleGribMessage:
         with open(filename, "wb") as f:
             pickle.dump(obj, f, protocol)
         # TODO: resolve this or remove the test.
-        # with open(filename, "rb") as f:
-        #     nobj = pickle.load(f)
-        # assert nobj == obj
+        #  with open(filename, "rb") as f:
+        #      nobj = pickle.load(f)
+        #  assert nobj == obj
 
     @pytest.mark.parametrize("protocol", TESTED_PROTOCOLS)
-    def test_message(self, protocol, messages):
+    def test_message(self, protocol, messages, tmp_path):
         obj = next(messages)
-        self.pickle_obj(obj, protocol)
+        self.pickle_obj(obj, protocol, tmp_path)
 
     @pytest.mark.parametrize("protocol", TESTED_PROTOCOLS)
-    def test_message_data(self, protocol, messages):
+    def test_message_data(self, protocol, messages, tmp_path):
         obj = next(messages).data
-        self.pickle_obj(obj, protocol)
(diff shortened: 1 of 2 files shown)


src/iris_grib/tests/unit/message/test_pickle_message.py
```diff
@@ -0,0 +1,48 @@
...
+    def messages(self):
+        path = tests.get_data_path(("GRIB", "fp_units", "hours.grib2"))
+        return GribMessage.messages_from_filename(path)
+
+    def pickle_obj(self, obj, protocol, tmp_path):
+        # NOTE: Neither GribMessage, nor its lazy ".data", read back from
+        # a pickled obj. Currently, this only checks that they can be pickled
+        # successfully.
+        filename = tmp_path / ".pkl"
+        with open(filename, "wb") as f:
```
trexfeathers, line 33: Pathlib offers even more convenience, FYI:

```suggestion
        with filename.open("wb") as f:
```

trexfeathers (approved):
Looks good, some neat use of PyTest best practice.

I know @pp-mo already signed off on the removal of Iris from these tests.

pp-mo:
Just a bit of a remaining worry here, I think.
Isn't the whole point of our interest in pickling that it is essential to the use of dask.distributed ?
In which case, the fact that you can't successfully reconstruct a cube based on a grib load is still a problem I think.
So I don't think we're properly fixed here.  TBC #

ESadek-MO:
> Just a bit of a remaining worry here, I think. Isn't the whole point of our interest in pickling that it is essential to the use of dask.distributed ? In which case, the fact that you can't successfully reconstruct a cube based on a grib load is still a problem I think. So I don't think we're properly fixed here. TBC #202

The load back had been excluded before I changed these tests. All I did is copy the code from Iris and add it to the TODO here, so that it was apparent what code was missing. If it had been fixed, these tests were outdated, and I have not edited the content of the code outside fo the pickle protocol and structure.

pp-mo:
> The load back had been excluded before I changed these tests .... If it had been fixed, these tests were outdated, and I have not edited the content of the code outside of the pickle protocol and structure.

OK very sorry 😢 I had not realised that !!

I had assumed that this got fixed at some point and so that #202 "should have been" closed !!

PR #637, opened 2025-04-22
Aligned manifest with scitools template.

pp-mo:
This aligns with the new template introduced in https://github.com/SciTools/.github/pull/131
(We probably want to wait until that is agreed)

I have tested this new version of MANIFEST.in, and the only change in the source distribution it produces is that it no longer packages `CODE_OF_CONDUCT.md`.
Which I think is actually no bad thing.


pp-mo:
Newly updated, due to changes in the template file.
This should now compare nicely with the [the latest template](https://github.com/SciTools/.github/blob/7363ee13d94e662d97d5a4e9f8166a09e5ccff88/templates/MANIFEST.in)

HGWright (approved):
Matches the new template, and Iris too.

Issue #574, opened 2024-11-22
Address disabled repo-review checks


pp-mo:
QA checks which we expect later to review + possibly remove.

Controls are pyproject.toml [here](https://github.com/SciTools/iris-grib/blob/main/pyproject.toml#L154-L166)

### Summary of controls
The ultimate goal is to have no 'ignores'.
There are no special config settings at present

### Remaining temporarily disabled checks 
  - [x] MY102 MyPy show_error_codes deprecated
  - [x] MY105 MyPy redundant-expr deprecated
  - [x] PC170 Uses PyGrep hooks
  - [ ] ~~PC180 Uses prettier~~
  - [x] PP309 Filter warnings specified

pp-mo:
I think #638 leaves only the 'prettier' recommendation to still be addressed

stephenworsley:
Scitools has broadly decided to opt out of prettier so I think we can close this. If we adopt it in the future, we can turn it on for each repo when that happens.

PR #638, opened 2025-04-24
Address repo-review checks

stephenworsley:
Addresses #574

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -104,3 +104,28 @@ repos:
     rev: "v0.24.1"
     hooks:
     - id: validate-pyproject
+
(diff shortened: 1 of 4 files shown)


pp-mo (approved):
This certainly looks like progress.

stephenworsley:
> > This certainly looks like progress.
> 
> ... but it still doesn't quite close #574. What is the situation with 'prettier' ? I understood we were blocked on using that locally (on the MetOffice desktop), but might that have changed recently ??

I've not not looked too much into the prettier issue, but it's the only remaining check on a bunch of SciTools repos now. I was under the impression it was either blocked or we had decided against it for the time being.

PR #677, opened 2025-07-22
Import and run grib-dependent attribute handling tests from Iris.

pp-mo:
N.B. this will not function until https://github.com/SciTools/iris/pull/6566 is merged,
since it uses the new tests provided there.

I think it **_should_** run once that is in, since we test against iris main branch.
Though we do also test against latest-release, and that will still fail until we update this...

### TODO... 
Made into draft as not currently viable...
  * [x] add pytest skips to ensure that tests pass even when the Iris version does not provide the newer tests to import

### DIFF
--- a/src/iris_grib/tests/integration/iris_integration/test_nc_attribute_handlers.py
+++ b/src/iris_grib/tests/integration/iris_integration/test_nc_attribute_handlers.py
@@ -1,0 +1,37 @@
+# Copyright iris-grib contributors
+#
+# This file is part of iris-grib and is released under the BSD license.
+# See LICENSE in the root of the repository for full licensing details.
+"""
+Integration tests for reading/writing GRIB_PARAM attributes in netcdf files.
+
+N.B. the relevant support code and classes (including testing) are all part of in Iris,
+along with the ones for "STASH" and "ukmo__process_flags" ones.
+
+But all the testing there is optional, and skipped if iris-grib is not installed, so that
+the Iris repo doesn't have to maintain iris-grib as a test dependency.
+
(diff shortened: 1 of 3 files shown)


pp-mo:
Think this proves itself now, since https://github.com/SciTools/iris/pull/6566 is merged, we now have an actual running test example
[ci-tests / tests py3.12 - Successful in 2m](https://github.com/SciTools/iris-grib/actions/runs/16475452945/job/46576457574?pr=677)

src/iris_grib/tests/integration/iris_integration/test_nc_attribute_handlers.py
```diff
@@ -0,0 +1,54 @@
...
+"""
+
+import pytest
+
+# Start by attempting the key imports from Iris, and skipping all tests if any fail.
+# Needed to run with older Iris (releases) prior to the tests being added.
+# Unfortunately we can't actually use the returned items, as dynamic typing upsets MyPy.
+pytest.importorskip("iris.tests.integration.netcdf.test_load_managed_attributes")
+pytest.importorskip("iris.tests.integration.netcdf.test_save_managed_attributes")
+pytest.importorskip("iris.tests.unit.fileformats.netcdf.attribute_handlers")
```
trexfeathers, line 25: This is a very neat pattern ❤️

trexfeathers (approved):
Thanks @pp-mo, firstly for your diligence of making sure we test the concept, and secondly for working hard to establish a pattern that I hope we can 'play forward' to other areas of Iris-grib and in other satellite repos too.

pp-mo:
Thanks @trexfeathers !

PR #705, opened 2025-10-07
Pre-commit: use `ruff-check` instead of `ruff`

trexfeathers:
Should fix the failures seen across SciTools in the latest pre-commit auto-update, and can be shared via templating.


trexfeathers:
More evidence: already implemented in GeoVista

https://github.com/bjlittle/geovista/pull/1726

Issue #489, opened 2024-06-18
Introduce an alternative GRIB1 loading using iris_grib.message.GribMessage instead


trexfeathers:
Some notes are in #56 

**Contract:** the tests introduced in #488 must always pass

This new behaviour should NOT be activated at this stage

The philosophy is that the loading behaviour will be switchable - choose between legacy and new behaviour - so musn't just replace the existing code.

PR #738, opened 2025-12-12
Refactor: move grib1 legacy code into its own submodule.

pp-mo:
For now, this is an "almost pure" refactor, to isolate the existing GRIB1 loading code into its own specific submodule.
The exception is that I already introduced a switching mechanism to select old/new grib1 loading (cf #490)

This is intended as a starting point for a re-implementation of GRIB1 load support, as in  #489.

### DIFF
--- a/src/iris_grib/tests/integration/load_convert/test_sample_file_loads.py
+++ b/src/iris_grib/tests/integration/load_convert/test_sample_file_loads.py
@@ -118,8 +118,6 @@ class TestBasicLoad(tests.IrisGribTest):
         )
         self.assertCML(cube, _RESULTDIR_PREFIX + ("reduced_ll_grib1.cml",))
 
-    # TODO: fix this - possibly just update the reference cml
-    @pytest.mark.skip(reason="known change in eccodes?")
     def test_reduced_gg_grib1(self):
         cube = iris.load_cube(
             Path(eccodes.codes_samples_path()) / "reduced_gg_ml_grib1.tmpl"


pp-mo:
## Note about doctest formatting
I uncovered some peculiar behaviour here, regarding how doctest setup and cleanup sections need to be formatted.

I am **now** proposing like this:
```
    .. setuptest::
[[N.B. blank line]]
        import iris_grib   [[N.B. without ">>>"]]

    .. doctest::

        >>> iris_grib.load(path)   [[N.B. with ">>>"]]
```
\-- mostly, just because that matches Iris usage.
This works for a separate doctest run via Sphinx (as in Iris).
However, this style does **not** work with pytest, where you must add ">>>" to the setup lines
And vice versa : that style does **not** work with Sphinx.

Prior to changes here, iris-grib was actually running all doctests **_twice_**, 
since the main 'tests' action used to run`pytest --pyargs iris_grib` -- which finds+runs docstring doctests as well.  
So, as soon as I introduced an in-docstring doctest using setup/clearup, I couldn't satisfy both pytest + Sphinx at once!!

In future I'd really **_like_** to adopt the [ncdata `run_doctests.py`](https://github.com/pp-mo/ncdata/blob/6e27328bae3082d97a3b7f29c9472e4638ba6fda/tools/run_doctests.py) doctest-runner style.
But unfortunately that **will** require the ">>>" everywhere (unlike Sphinx) -- since it runs on the bare un-built source files
(which is a large part of the point).
BTW, even though a big benefit of adopting the `run_doctests.py` would be that we _don't_ need a Sphinx build to run doctests,
unfortunately I think we **_must_** still use  the 'sphinx.ext.autodoc' extension or Sphinx can't cope with doctest sections at all, and will swallow them + not render the main example code -- either that, or it renders all the setup sections too :weary:

docs/conf.py
```diff
@@ -13,13 +13,8 @@
...
 
 from importlib.metadata import version as get_version
-import sys
-import os
+from pathlib import Path
 
-# If extensions (or modules to document with autodoc) are in another directory,
-# add these directories to sys.path here. If the directory is relative to the
-# documentation root, use os.path.abspath to make it absolute, like shown here.
-sys.path.append(os.path.abspath(".."))
```
trexfeathers: I believe this is a standard placeholder that Sphinx provides, to make it as easy as possible to write custom extensions in future. Did you remove it because you don't think we'll need it any time soon?

src/iris_grib/__init__.py
```diff
@@ -43,113 +39,135 @@
...
-    """A reference to the data payload of a single Grib message."""
-
-    __slots__ = ("dtype", "offset", "path", "shape")
-
-    def __init__(self, shape, dtype, path, offset):
-        self.shape = shape
-        self.dtype = dtype
-        self.path = path
-        self.offset = offset
+class Grib1LoadingMode(threading.local):
```
trexfeathers: We should probably have test coverage for this object, right?

noxfile.py
```diff
@@ -285,8 +285,7 @@ def tests(session: nox.sessions.Session, iris_source: str):
 
     run_args = [
         "pytest",
-        "--pyargs",
-        "iris_grib",
+        "src/iris_grib/tests",  # NB *don't* trigger doctests: they are done separately
```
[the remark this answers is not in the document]
trexfeathers: Edit: missed this comment: https://github.com/SciTools/iris-grib/pull/738#issuecomment-3657187405

docs/conf.py
pp-mo: I just thought it had come from Iris, where it probably used to be active + of value.
If it is part of the "standard" form, I would expect it to be commented out by default?

trexfeathers: After conversation offline: we're going with the more slimmed-down approach.

src/iris_grib/tests/integration/load_convert/test_sample_file_loads.py
```diff
@@ -118,6 +118,8 @@ def test_reduced_ll(self):
         )
         self.assertCML(cube, _RESULTDIR_PREFIX + ("reduced_ll_grib1.cml",))
 
+    # TODO: fix this - possibly just update the reference cml
+    @pytest.mark.skip(reason="known change in eccodes?")
```
trexfeathers: Is it possible that your branch is just out of date? I updated the CML in #716

pp-mo: It is possible that there is a weird intermittent problem here.  The data signatures (new-style) don't match.
If I comment out the skip, I get ...
```
E   AssertionError: CML do not match: /home/users/patrick.peglar/git/iris_grib/src/iris_grib/tests/results/integration/load_convert/sample_file_loads/reduced_gg_grib1.cml
E   --- Reference
E   +++ Test result
E   @@ -43 +43 @@
E   -    <data checksum="0x514a7763" dtype="float64" shape="(6114,)"/>
E   +    <data checksum="0x6fe0ee2c" dtype="float64" shape="(6114,)"/>
```

trexfeathers: OK so the branch is up to date; what about the environment you are using for local testing? It needs to be ecCodes version `2.44.0` or higher.

trexfeathers (approved):

PR #767, opened 2026-03-23
Fix rounding of scaled integers.

### LINKED ISSUE #686: Saved coordinate values are different with numpy2.
Coordinate values in a saved grib file changed when numpy v1.26.4 was replaced by v2.2.6

The example has coordinates are derived from a PP file, with the following original values:
```
FLD:  lbrow 1920
FLD:  lbnpt 2560
FLD: bzx -0.070312500000000000000000000000
FLD: bdx 0.140625000000000000000000000000
FLD: bzy -90.046875000000000000000000000000
FLD: bdy 0.093750000000000000000000000000
```
this loads as XY coordinates (unbounded) with the following endpoints:
```
Lons: 0.0703125 ... 359.9297
Lats: -89.953125 ... 89.953125
```
and the following step values:
```
Lon step value: 0.140625
Lat step value: 0.09375
```

The line here converts these to these key values in the saved file:
  * iDirectionIncrement = 140625
  * jDirectionIncrement = 93750
  * longitudeOfFirstGridPoint = 70312
  * longitudeOfLastGridPoint = 3599296**87** (numpy 1.26) --> 3599296**96** (numpy 2.2)
  * latitudeOfFirstGridPoint = -8995312**5** (numpy 1.26) --> -8995312**8** (numpy 2.2)
  * latitudeOfLastGridPoint = 8995312**5**  (numpy 1.26) --> 8995312**8** (numpy 2.2)

The reason for the differences is that numpy2 doesn't promote value types in the same way.
The actual calculation is [here](https://github.com/SciTools/iris-grib/blob/v0.21.0/src/iris_grib/_save_rules.py#L184-L209)
Notably, this does effectively `int(value * 1000000)`.
But, numpy2 reproduces the previous values if we instead use `int(value.astype("f8") * 1000000)`

**However** the original different values caused an error in other software, which was apparently checking that, e.g. 
`longitudeOfLastGridPoint  == longitudeOfFirstGridPoint + iDirectionIncrement * (Ni - 1)`
The above calculation fix doesn't necessarily guarantee that, especially as our source is 32-bit PP data, which has caused us previous problems with the rounding of coordinate values.

So possibly, the fix needs to ensure the expected relationship between (start, stop, step, number), when these are derived as integer values from coordinate point values.

pp-mo:
Closes #686 

NOTE: I've checked that the new test fails with the older code, producing the same values noted in the report :
```
src/iris_grib/tests/unit/save_rules/test_latlon_first_last.py:31 (Test.test_latlons[wrapped])
[-89953128, 89953128, 70312, 359929696] != [-89953125, 89953125, 70312, 359929687]

Expected :[-89953125, 89953125, 70312, 359929687]
Actual   :[-89953128, 89953128, 70312, 359929696]
```

### DIFF
--- a/src/iris_grib/_save_rules.py
+++ b/src/iris_grib/_save_rules.py
@@ -219,28 +219,22 @@ def latlon_first_last(x_coord, y_coord, grib):
     if x_coord.has_bounds() or y_coord.has_bounds():
         warnings.warn("Ignoring xy bounds")
 
-    # XXX Pending #1125
-    #    eccodes.codes_set_double(grib, "latitudeOfFirstGridPointInDegrees",
-    #                            float(y_coord.points[0]))
-    #    eccodes.codes_set_double(grib, "latitudeOfLastGridPointInDegrees",
-    #                            float(y_coord.points[-1]))
-    #    eccodes.codes_set_double(grib, "longitudeOfFirstGridPointInDegrees",
-    #                            float(x_coord.points[0]))
-    #    eccodes.codes_set_double(grib, "longitudeOfLastGridPointInDegrees",
-    #                            float(x_coord.points[-1]))
-    # WORKAROUND
-    eccodes.codes_set_long(
-        grib, "latitudeOfFirstGridPoint", int(y_coord.points[0] * 1000000)
-    )
-    eccodes.codes_set_long(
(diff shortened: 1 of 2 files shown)


trexfeathers (approved):
Thanks @pp-mo. NumPy arithmetic that includes a Python scalar is the most common NumPy v2 breaking change we've been seeing 👍

PR #768, opened 2026-03-23
Refactor docs and add new example

### LINKED ISSUE #657: Breaking change introduced between v0.17 and v0.21
We have recieved files which used to load properly in an environment with v0.17 and fail to load in v0.21. The error raised is:
`"Product definition section 4 has missing scaled value of second fixed surface"`
For another file, loading succeeds but happens differently.

Potential causes for this could be #318 or perhaps an update to eccodes.

trexfeathers:
Closes #657

As usual: I recommend reviewing commit-by-commit!


pp-mo (changes requested):
I think this is just utterly brilliant!
It really shows how just re-organising as-per Diataxis adds clarity.
-- the new content page is great too

I'm just going to question whether it's _worth_ controlling the lhs sidebar, 
because it seems to me that the default is ~ok, and I think new behaviour is not totally sorted

docs/how_to/simple_grib_io.rst
```diff
@@ -0,0 +1,29 @@
+Simple GRIB Loading and Saving with Iris
```
pp-mo:
```suggestion
.. _simple_load_save:

Simple GRIB Loading and Saving with Iris
```

pp-mo (changes requested):
Tiny fixes, mostly for cross-linkage

pp-mo (approved):
:+1:

PR #770, opened 2026-03-24
Update Python and Iris minimum pins

### LINKED ISSUE #465: iris min pin
Reference #463.

Enforce an `iris>=3.10` for `iris_grib` when `iris` 3.10 is released.

Also purge the defensive code `iris_grib.__init__._aslazydata_has_meta` along with an appropriate refactor.

trexfeathers:
- Closes #465
- Updates Python versions to align with [SPEC 0](https://scientific-python.org/specs/spec-0000/)


pp-mo (approved):
Ok I think this is good

PR #771, opened 2026-03-24
Release notes for v0.22

trexfeathers:
Did my best with the internal stuff too since it shows what it takes to keep a repository going


pp-mo (changes requested):
One definite glitch + a couple of possible additions?

pp-mo (approved):

PR #777, opened 2026-03-31
Update RTD server 

pp-mo:
Since RTD is now warning that ubunutu-20 is deprecated.

To highest current ubuntu version = 26.04
_Not_ using 'latest', for stability reasons.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -3,7 +3,7 @@ version: 2
 build:
   os: ubuntu-24.04
   tools:
-    python: "mambaforge-4.10"
+    python: mambaforge-4.10
   jobs:
     post_checkout:
       # The SciTools/iris-grib repository is shallow i.e., has a .git/shallow,
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,7 +1,7 @@
 version: 2
 
 build:
-  os: ubuntu-26.04
+  os: ubuntu-24.04
   tools:
     python: "mambaforge-4.10"
   jobs:
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -3,7 +3,7 @@ version: 2
 build:
   os: ubuntu-26.04
   tools:
-    python: mambaforge-4.10
+    python: "mambaforge-4.10"
   jobs:
     post_checkout:
       # The SciTools/iris-grib repository is shallow i.e., has a .git/shallow,
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,7 +1,7 @@
 version: 2
 
 build:
-  os: ubuntu-20.04
+  os: ubuntu-26.04
   tools:
     python: mambaforge-4.10
   jobs:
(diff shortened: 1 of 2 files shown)


pp-mo:
Modded to see if v24 will work, as v26 apparently does not.

More likely, it is time to remove reference to mamba ???

ukmo-ccbunney (commented):
Looks fine, apart from failing test (seems like a timeout?)
Just one comment.

pp-mo:
> Looks fine, apart from failing test (seems like a timeout?) Just one comment.

It isn't a test timeout.  It's a url that is not responding.
Unfortunately there's a lot of carnage out there ATM

What was your other issue? I think you didn't post the comment

.readthedocs.yml
```diff
@@ -1,9 +1,9 @@
 version: 2
 
 build:
-  os: ubuntu-20.04
+  os: ubuntu-24.04
   tools:
-    python: mambaforge-4.10
+    python: "mambaforge-4.10"
```
ukmo-ccbunney: We're the quotes necessary for something, or just from testing?

pp-mo:
> It isn't a test timeout. It's a url that is not responding. Unfortunately there's a lot of carnage out there ATM

Actually I'm probably wrong about that.
Re-spinning

ukmo-ccbunney (commented):
Sorry - didn't post the comment!

(Yes - I meant that the URL access (repo-review) is timing out on the tests, not the tests themselves).

.readthedocs.yml
pp-mo: I was just following the docs, but yes they probably aren't necessary.
It happened because I tried replacing this with various other things to get around failure with ubuntu-26.04
But nothing worked so I put it back to ubuntu-24 and "mambaforge".
TBH I don't really understand what these options really mean : there is a clear list of alternatives in the docs, but what it is actually affecting seems rather vague!

pp-mo: > I tried replacing this with various other things to get around failure with ubuntu-26.04 But nothing worked 

This an ongoing problem : I couldn't make it work any way with ubuntu-26.
Could be a problem for the future

ukmo-ccbunney (changes requested):

.readthedocs.yml
```diff
@@ -1,7 +1,7 @@
 version: 2
 
 build:
-  os: ubuntu-20.04
+  os: ubuntu-24.04
   tools:
     python: mambaforge-4.10
```
ukmo-ccbunney, line 6: According to [the RTD docs,](https://docs.readthedocs.com/platform/stable/config-file/v2.html#build-tools-python), `mambaforge-4.10` is deprecated too...

ukmo-ccbunney (changes requested):

pp-mo:
## Update 2026-07-29
Rebased.
~Back in **Draft**, while I look (briefly?) into a better resolution of the options.~

updated to latest os="ubuntu" + python="miniforge3" (but version pinned),
seems to work

trexfeathers (approved):
Can't argue with a successful docs build!

PR #795, opened 2026-05-26
Switch to the standard Lychee link check

trexfeathers:
Based on this template, which is proven in Iris:

https://github.com/SciTools/.github/blob/main/templates/github/workflows/ci-linkchecks.yml

If you would like a demonstration, [here is an action on my fork](https://github.com/trexfeathers/iris-grib/actions/runs/26439148924/job/77828999843). The only failed link is the one for this PR, since it did not exist yet at the time of running.


trexfeathers:
>...
>This PR includes changes that may be worth sharing via templating.
>...

https://github.com/SciTools/.github/pull/265

pp-mo (changes requested):
All the changes look great.
Re-checked against the template ✅ 

But I also did a global search + found that there is still code for the Sphinx link-checking in
`docs/Makefile` and `docs/make.bat`.
So is it worth removing those ?

pp-mo (approved):

Issue #538, opened 2024-09-10
Adopt common code for setting "scalingFactor" and "value" keys


pp-mo:
As for example #205, but some general functionality.
Probably something like `scalefactor_and_value(value, min_factor=0, max_factor=7)`

A fair bit of work to apply this everywhere, naturally.
But we use currently use fixed values almost everywhere, which is just a limitation waiting to bite someone.

pp-mo:
See also : #211

PR #813, opened 2026-06-24
Gdt40 save

### LINKED ISSUE #785: Support saving to GDT 40
Currently iris-grib supports loading GDT 40 (see [here](https://github.com/SciTools/iris-grib/blob/0e53c681e9b4c1f67c6e1291cd9b711492730df1/src/iris_grib/_grib2_convert.py#L1076)) but not saving it. 

We would like to write out data on an octahedral reduced gaussian grid.

I can provide some example files of what "good" looks like.

pp-mo:
Addresses #785 

So far, so highly sketchy + experimental

~NOTE: I don't really like https://github.com/SciTools/iris-grib/pull/813/commits/94a3d38b7e2936e65cb106e9fdafaf817211facc because we **only** want to support calling without coords for the sake of the existing unit tests.~
~But I tried adapting some of the existing mock-ist tests + it's really quite awkward~


pp-mo:
## Update 2026-07-21
* Successfully saves to GDT40, and reads back as matching cubes (though not specifically tested)
* all tests passing
* reviewed content for sanity, added/fixed comments
* checked that GRIB1 gaussian data is also compatible, and will save by adding a `GRIB2_GRID_TEMPLATE = 40`
   (but we don't have any in iris-test-data)

Not implemented:
* ability to save suitable *regular* gaussian data to GDT40 
  * currently will save, but as 'irregular lat-lon' which is [GDT4](https://github.com/SciTools/iris-grib/blob/3b89c6cd16c2378e4442c8ea2ca559200d4a43b3/src/iris_grib/_save_rules.py#L985)
  * ideally, would respect the `GRIB2_GRID_TEMPLATE` to enable this?
* check that the parallel values in the cube exactly match what "would" be the [values for the resulting GDT40 grib message](https://github.com/SciTools/iris-grib/blob/3b89c6cd16c2378e4442c8ea2ca559200d4a43b3/src/iris_grib/_save_rules.py#L949-L954) 

Still to-do:
  * [ ] add documentation and whatsnew coverage
        **update** see https://github.com/pp-mo/iris-grib/pull/3
  * [ ] add a unit test for gdt40 save routine - e.g. a test_grid_definition_template_40.py, like [this one](https://github.com/SciTools/iris-grib/blob/v0.22.0/src/iris_grib/tests/unit/save_rules/test_grid_definition_template_30.py) 
      **update** see https://github.com/pp-mo/iris-grib/pull/4

src/iris_grib/__init__.py
```diff
@@ -349,15 +351,35 @@ def save_pairs_from_cube(cube):
...
-        grib_message = eccodes.codes_grib_new_from_samples("GRIB2")
-        _save_rules.run(slice2D, grib_message, cube)
+    slice_param = [y_coord, x_coord] if dim_coords else [x_coord]
+    for slice2D in cube.slices(slice_param):
+        if _save_rules.is_grid_definition_template_40(cube, x_coord, y_coord):
+            template_name = "reduced_gg_sfc_grib2"
+        else:
+            template_name = "GRIB2"
+        grib_message = eccodes.codes_grib_new_from_samples(template_name)
+        result = _save_rules.run(slice2D, grib_message, cube, x_coord, y_coord)
```
trexfeathers: `_save_rules.run()` does not return anything.

src/iris_grib/_grib1_legacy/grib1_load_rules.py
```diff
@@ -168,6 +177,17 @@ def grib1_convert(grib):
         )
         units = "???"
 
+    # Special!!
+    if grib.table2Version == 128 and grib.indicatorOfParameter == 134:
```
trexfeathers: @pp-mo has a concern that this is happening too broadly. I think we had a way to make it only happen when a hybrid height is involved?

src/iris_grib/_grib2_convert.py
```diff
@@ -1223,6 +1225,18 @@ def grid_definition_template_40_reduced(section, metadata, cs):
...
+    #  the unstructured dimension, so they can match their dimensions.
+    metadata["dim_coords_and_dims"].append(
+        (
+            DimCoord(
+                np.arange(x_points.shape[0]), long_name="gaussian_grid", units="1"
+            ),
+            0,
+        )
+    )
+
```
trexfeathers, line 1239: Do we need to confirm that we are definitely in a hybrid height situation, or is that guaranteed when it's template 40?

src/iris_grib/_save_rules.py
```diff
@@ -826,31 +827,171 @@ def grid_definition_template_140(cube, grib):
...
+    if ok:
+        if not lats_increasing:
+            # Put the set of latitude values into the expected order
+            lat_vals = lat_vals[::-1]
+        lons = x_coord.points
+        lons_increasing = np.diff(lons[lats == lat_vals[0]]).min() >= 0
+
+        prev_maxind = -1
+        for this_lat in lat_vals:
+            thislat_inds = np.where(lats == this_lat)[0]
```
trexfeathers, line 892:
```suggestion
            (thislat_inds,) = np.where(lats == this_lat)
```

This caught me out wondering "why does he only want the first one?". So I thought unpacking would be clearer.

src/iris_grib/_save_rules.py
```diff
@@ -826,31 +827,171 @@ def grid_definition_template_140(cube, grib):
...
+    return ok
+
+
+def grid_definition_template_40(cube, grib, x_coord, y_coord):
+    # For the gaussian grids, for now it appears that to function the message must be
+    # based on a different template.
+    # grib = eccodes.codes_grib_new_from_samples("reduced_gg_sfc_grib2")
+    # --OR-- possibly also, can set the "computed key" 'gridType = gg' ??
+    #
+    # **FOR NOW:** this is handled as a special case by the caller:
```
trexfeathers: I can't tell if the word `this` is referring to the previous comment sentences, or referring to template 40, or what. I am aware of `is_grid_definition_template_40()`, but I'm not sure what I'm supposed to take from this comment.

src/iris_grib/_save_rules.py
```diff
@@ -826,31 +827,171 @@ def grid_definition_template_140(cube, grib):
...
+    #  'grid_definition_section' --> 'is_grid_definition_template_40'
+
+    # NB some code here is also duplicated from the 'is..' routine : TODO improve DRY??
```
trexfeathers: Agree. It looks like the information you need is `lats_increasing` and `lons_increasing`? Seems appropriate for factoring out.

src/iris_grib/_save_rules.py
```diff
@@ -826,31 +827,171 @@ def grid_definition_template_140(cube, grib):
...
+    eccodes.codes_set_missing(grib, "iDirectionIncrement")
+
+    lon_repeats = [np.sum(lats == onelat) for onelat in lat_vals]
+    eccodes.codes_set(grib, "numberOfOctectsForNumberOfPoints", 2)
+    eccodes.codes_set(grib, "interpretationOfNumberOfPoints", 1)
+    eccodes.codes_set_array(grib, "pl", list(lon_repeats))
+
+    # # Check the latitude values calculation
+    # # ??? but apparently, this can't be called in this way ???
+    # y_points = eccodes.codes_get_array(grib, "distinctLatitudes")
```
trexfeathers: What's the plan with this?

src/iris_grib/_save_rules.py
```diff
@@ -826,31 +827,171 @@ def grid_definition_template_140(cube, grib):
...
+
+    eccodes.codes_set(grib, "longitudeOfFirstGridPoint", int(1.0e6 * lon0))
+    eccodes.codes_set(grib, "longitudeOfLastGridPoint", int(1.0e6 * lon1))
+
+    eccodes.codes_set(grib, "N", n_lat_vals // 2)
+    eccodes.codes_set(grib, "Nj", n_lat_vals)
+    eccodes.codes_set_missing(grib, "Ni")
+
+    eccodes.codes_set_long(grib, "basicAngleOfTheInitialProductionDomain", 0)
+    eccodes.codes_set_long(grib, "resolutionAndComponentFlags", 0)
```
trexfeathers, line 954: Why are we not involving `_RESOLUTION_AND_COMPONENTS_GRID_WINDS_BIT` here?

src/iris_grib/__init__.py
```diff
@@ -349,15 +351,35 @@ def save_pairs_from_cube(cube):
...
+        if result is not None:
+            grib_message = result
```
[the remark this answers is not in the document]
pp-mo: :+1: This is legacy from an older approach.
Removed

src/iris_grib/_save_rules.py
```diff
@@ -826,31 +827,171 @@ def grid_definition_template_140(cube, grib):
...
+
+    # We want to do the equivalent of 'horizontal_common', but we can't call that as it
+    #  calls "grid_dims" to set Ni/Nj, which we need to control differently.
+    # So we here call its other workers : "shape_of_the_earth" + "scanning_mode_flags"
+    shape_of_the_earth(cube, grib, cs=x_coord.coord_system)
+    scanning_mode_flags(x_coord, y_coord, grib)
+
+    eccodes.codes_set(
+        grib, "latitudeOfFirstGridPoint", int(1.0e6 * lats[0])
+    )  # in micro degrees
```
trexfeathers, line 938: I understand from the [GRIB note](https://github.com/wmo-im/GRIB2/blob/ad61905de4570eaf4f3ab1a1df3a07122bf4d338/notes/Template_notes.csv#L10) [^1] that we are dealing with "ordinary cases" only? This would explain the deviation from existing template functions (which are responsive to `_coord_system` and `_DEFAULT_DEGREE_UNITS`).

[^1]: Basic angle of the initial production domain and subdivisions of this basic angle are provided to manage cases where the recommended unit of 10-6 degrees is not applicable to describe the longitudes and latitudes, and direction increments. For these descriptors, the unit is equal to the ratio of the basic angle and the subdivisions number. For ordinary cases, zero and missing values should be coded, equivalent to the respective values of 1 and 106 (10-6 degrees unit).

src/iris_grib/_save_rules.py
```diff
@@ -826,31 +827,171 @@ def grid_definition_template_140(cube, grib):
...
+        lon0, lon1 = lons.max(), lons.min()
+
+    eccodes.codes_set(grib, "longitudeOfFirstGridPoint", int(1.0e6 * lon0))
+    eccodes.codes_set(grib, "longitudeOfLastGridPoint", int(1.0e6 * lon1))
+
+    eccodes.codes_set(grib, "N", n_lat_vals // 2)
+    eccodes.codes_set(grib, "Nj", n_lat_vals)
+    eccodes.codes_set_missing(grib, "Ni")
+
+    eccodes.codes_set_long(grib, "basicAngleOfTheInitialProductionDomain", 0)
```
trexfeathers, line 953: I understand from the [GRIB note](https://github.com/wmo-im/GRIB2/blob/ad61905de4570eaf4f3ab1a1df3a07122bf4d338/notes/Template_notes.csv#L10) [^1] that we are dealing with "ordinary cases" only? This would explain the `0` and missing values for `basicAngleOfTheInitialProductionDomain` and `iDirectionIncrement`.

[^1]: Basic angle of the initial production domain and subdivisions of this basic angle are provided to manage cases where the recommended unit of 10-6 degrees is not applicable to describe the longitudes and latitudes, and direction increments. For these descriptors, the unit is equal to the ratio of the basic angle and the subdivisions number. For ordinary cases, zero and missing values should be coded, equivalent to the respective values of 1 and 106 (10-6 degrees unit).

src/iris_grib/_save_rules.py
```diff
@@ -826,31 +827,171 @@ def grid_definition_template_140(cube, grib):
...
+    eccodes.codes_set_long(grib, "resolutionAndComponentFlags", 0)
+    eccodes.codes_set_missing(grib, "iDirectionIncrement")
+
+    lon_repeats = [np.sum(lats == onelat) for onelat in lat_vals]
+    eccodes.codes_set(grib, "numberOfOctectsForNumberOfPoints", 2)
+    eccodes.codes_set(grib, "interpretationOfNumberOfPoints", 1)
```
trexfeathers, line 959: https://github.com/wmo-im/GRIB2/blob/ad61905de4570eaf4f3ab1a1df3a07122bf4d338/GRIB2_CodeFlag_3_11_CodeTable_en.csv#L3

👍

src/iris_grib/_save_rules.py
```diff
@@ -826,31 +827,171 @@ def grid_definition_template_140(cube, grib):
...
+    eccodes.codes_set(grib, "N", n_lat_vals // 2)
+    eccodes.codes_set(grib, "Nj", n_lat_vals)
+    eccodes.codes_set_missing(grib, "Ni")
+
+    eccodes.codes_set_long(grib, "basicAngleOfTheInitialProductionDomain", 0)
+    eccodes.codes_set_long(grib, "resolutionAndComponentFlags", 0)
+    eccodes.codes_set_missing(grib, "iDirectionIncrement")
+
+    lon_repeats = [np.sum(lats == onelat) for onelat in lat_vals]
+    eccodes.codes_set(grib, "numberOfOctectsForNumberOfPoints", 2)
```
trexfeathers, line 958: I can't find the official reference but I understand what you have done based on this ECMWF table:

https://codes.ecmwf.int/grib/format/grib2/sections/3/

👍

trexfeathers (dismissed):

src/iris_grib/__init__.py
pp-mo: Yes, legacy of a former change.
Copilot spotted that + I already removed it.

src/iris_grib/_grib1_legacy/grib1_load_rules.py
pp-mo: Actually I think this relates to some **_other_** temporary test data that Laura posted me for evaluation of the GDT40 message in it.  And in that, the surface pressure fields were in an ecmwf-specific GRIB1 form.
I think it is not relevant to present needs -- the "aifs" data has GRIB2 surface pressure fields.
So I will remove this altogether...

pp-mo: https://github.com/SciTools/iris-grib/pull/813/commits/215e82e1f1657349ea34cdddc0191c6c19f10c73

src/iris_grib/_grib2_convert.py
pp-mo: No, I think it's fine to provide this dim-coord, _whenever_ we load reduced-GDT40 data.
It is [these lines](https://github.com/pp-mo/iris-grib/blob/83b77a03c55fb936b510863aea93f788d2430be5/src/iris_grib/_grib2_convert.py#L1694-L1752) which trigger the (attempted) creation of a hybrid vertical coord -- none of which is changed here. 
The change here just makes it _work_ when it tries it, because it needs the relevant dimension to be identifiable.

We couldn't load this type of data before **at all**, so no backwards-compatibility concerns.

src/iris_grib/_save_rules.py
pp-mo: I've made this clearer (probably!)

TBH this still upsets me.  There is an alternative, but it seems to require us to use the computed key:
 `codes_set("gridType", "gg")` instead of `codes_set("gridDefinitionTemplateNumber", 40)`
 see [notes here](https://github.com/SciTools/iris-grib/issues/785#issuecomment-4801778654).
 
 The ugly hack is just my pragmatic way to avoid adding another computed-key dependency.

src/iris_grib/_save_rules.py
pp-mo: This is just the default value, because we are not supporting 'grid winds', as that is a different template:
"Grid definition template 3.41 – rotated Gaussian latitude/longitude"

src/iris_grib/_save_rules.py
pp-mo: To some extent, **this is why Laura pointed me at some "real data" of another user containing GDT40 fields**
(which came from ecmwf).
And so I have just checked that the **various "other" keys in the dump match what iris-grib is now producing.**

Though it is also true that in various cases, as you point out above, we don't supporting alternative scaling or resolution controls. 
Ideally, this kind of thing needs to be better handled globally, but [that is a TTD](https://github.com/SciTools/iris-grib/issues/538). 

I've taken the pragmatic view that it is just 'not needed' for MVP in this case.  The practical cases we are dealing with so far don't seem to ever need it.

src/iris_grib/_save_rules.py
pp-mo: Actually I'm **_more_** bothered by the simple possibility of a non-standard 'first' latitude (or longitude) value.
There doesn't seem to me to be enough detail in the spec to nail down how such things should be interpreted (or encoded), though you could use computed keys (latitudes / longitudes) which I guess provides access to how eccodes thinks it should be done.

Achieving more diligence on this is just hard :cry:

src/iris_grib/_save_rules.py
sawom666: Sorry for being pedantic...
There is no gridType of "gg": we have "reduced_gg", "regular_gg" etc

trexfeathers: Appreciate your involvement @sawom666!

src/iris_grib/_save_rules.py
trexfeathers: What we have is clearly a legitimate approach, even if there are some potential edge cases. The most I'd support would be raising an exception to avoid those situations you are worried about. Anything beyond seems like too much.

src/iris_grib/_save_rules.py
pp-mo: > Sorry for being pedantic... There is no gridType of "gg": we have "reduced_gg", "regular_gg" etc

Sorry that was just from memory. In this case, I definitely mean "reduced_gg" !

src/iris_grib/_save_rules.py
pp-mo: My problem is that I don't have a means of deriving the correct/expected value of "latitudeOfFirstGridPoint" either
-- it depends on the number of latitudes, like the other latitude values.
So as with those others, for now I am just trusting that the provided value is correct.

I have tested this with data read from GDT40 (i.e. it roundtrips), and also with similar reduced-gg data (with a different 'N') loaded from GRIB1 fields.  In those cases, the first and last longitude+latitude points are correct
 -- that is, they equate on round-tripping, which validates the save code in those cases.
 And, originally I had the last longitude value wrong (effectively, lons[-1] instead of lons.max()), and it did **_not_** round-trip until I fixed that.

It's possible that the initial+final **latitude** values don't actually matter : they seem to be effectively _redundant_, due to the full set of latitude values being determined by the gaussian lats calc (and 'N').  
At least, IIUC, there is no practical way for the range to be other than the whole globe, so recording the first+last values independently is **probably** pointless.  
It's a bit of a typical GRIB format conundrum : the definition has both ambiguity + redundancy!

I think we should maybe discuss this offline ...

src/iris_grib/_save_rules.py
pp-mo: Yes I think so, as for the previous point.

src/iris_grib/_save_rules.py
pp-mo: OK I hope that https://github.com/SciTools/iris-grib/pull/813/commits/c2558596b5f93b242a52f9951ed81e5ccc485521 now solves this,
since it is now checking that the latitudes derived from the message config exactly match those in the original cube

src/iris_grib/_save_rules.py
pp-mo: https://github.com/SciTools/iris-grib/pull/813/commits/c2558596b5f93b242a52f9951ed81e5ccc485521 fixes it

pp-mo:
Ok since the policy change in https://github.com/SciTools/iris-grib/pull/813/commits/c2558596b5f93b242a52f9951ed81e5ccc485521 I think the core solution looks a lot more sensible, and the ability to check against lats+lons derived from the message makes it much more trustworthy.

As well as no longer using `is_grid_definition_template_40` to choose a template in the top level  `save_pairs_from_cube`, I have also radically reduced it's scope to just
  \-  "1-d x and y coords on the same dimension, and recorded template-number = 40".  
So it scarcely needs a separate test.
Unfortunately that rather drastically affects the tests in [their separate PR](https://github.com/pp-mo/iris-grib/pull/3)
(but I have updated that now)

For both of those follow-on PRs ([tests](https://github.com/pp-mo/iris-grib/pull/3) and [docs](https://github.com/pp-mo/iris-grib/pull/4)), 
for now I've kept them separate, rebased them onto this so you can see what's going on, and am addressing the concerns already raised.  I think it's worth keeping them separate to handle the issues @trexfeathers raised there.
There shouldn't then be any drama merging those here, once we've resolved the outstanding issues already raised
(and before merging this)

src/iris_grib/_save_rules.py
pp-mo: I've done this now, but in the subsequent https://github.com/pp-mo/iris-grib/pull/3

src/iris_grib/_save_rules.py
ukmo-ccbunney: Resolving this thread as I can see you have switched to using the `reduced_gg` computed key in https://github.com/SciTools/iris-grib/commit/c2558596b5f93b242a52f9951ed81e5ccc485521

src/iris_grib/_save_rules.py
ukmo-ccbunney: The check of the lats/lons against the computed values looks sensible.
Out of interest, how did you choose the tolerance, e.g.
```python
tolerance = abs(lon1 - lon0) * 0.02 / n_values
```

src/iris_grib/_save_rules.py
ukmo-ccbunney: Resolving as outdated.

src/iris_grib/_save_rules.py
pp-mo: Basically the idea here is "2% of the average difference between adjacent values", since the values move in roughly equal steps, though changing faster near the poles.
We could certainly do this in other ways -- maybe a fixed absolute tolerance like 0.01 degrees ?

ukmo-ccbunney: No - happy with that. I was just wondering where the 2% came from (if it was documented somewhere in eccodes).

pp-mo: Background :  
from inspecting actual cases, I think the **latitude** value range is always symmetrical, "close" to -90..+90, and "close" to regular steps -- except they change more rapidly near +/-90.  The min/max get closer to +/-90 with increasing 'N'.
'N' is half the number of latitude points -- so practically a few * 10 up to a few * 1000 I guess.

For longitudes, the values are regularly spread over 0..360, but the numbers of points varies with the latitude -- practically can be as low as 4, again up to maybe 1000's.  

However the longitude values are simpler to understand/calculate + probably more stable (though I'm not clear how the start/end values can be shifted around the circle).
My concern with the lats is that the derivation from solving a D.E. could result in different values due to rounding/accuracy shifts on different machines.  So I don't trust them to be as repeatable.

src/iris_grib/_save_rules.py
```diff
@@ -826,31 +827,133 @@ def grid_definition_template_140(cube, grib):
...
+    eccodes.codes_set(grib, "interpretationOfNumberOfPoints", 1)
+    eccodes.codes_set_array(grib, "pl", list(lon_repeats))
+
+    (n_values,) = x_coord.shape  # coords are always 1-D
+    # Setting these 2 additional keys makes the message fully valid,
+    #  allowing us to read the 'correct' gaussian latitude and longitude values
+    eccodes.codes_set(grib, "numberOfValues", n_values)
+    eccodes.codes_set(grib, "numberOfDataPoints", n_values)
```
trexfeathers, line 966: Ref: https://github.com/SciTools/iris-grib/issues/785#issuecomment-4856544952

👍

pp-mo: Yes, I will credit @sawom666 by adding a mention in the whatsnew

pp-mo: https://github.com/pp-mo/iris-grib/pull/4/commits/db79916aaf000db61e8dae1470fa6bd85747fec8

docs/reference/release_notes.rst
```diff
@@ -9,6 +9,50 @@ What's new in iris-grib v0.23.0
...
+  to record source GRIB2 grid template numbers on load as a
+  ``GRIB2_GRID_TEMPLATE`` cube attribute.  This attribute is also used to control saving
+  of 1-D data to a "reduced gaussian" grid, and may be used for further save controls in
+  future.  See :ref:`template_record` for more information.
+  `(ISSUE#785) <https://github.com/SciTools/iris-grib/issues/785>`_,
+  `(PR#813) <https://github.com/SciTools/iris-grib/pull/813>`_
+
+* `@pp-mo <https://github.com/pp-mo>`_ , `@sawom666 <https://github.com/sawom666>`_,
+  `@trexfeathers <https://github.com/trexfeathers>`_,
+  `@chrisbunney <https://github.com/chrisbunney>`_ and
```
ukmo-ccbunney: It's **a** Chris Bunney, but not me!

```suggestion
  `@ukmo-ccbunney <https://github.com/ukmo-ccbunney>`_ and
```

ukmo-ccbunney (changes requested):
One change - I did raise this before, but seems to have got lost.

ukmo-ccbunney (approved):
I think we're all good, to the limit of my understanding!
Well done @pp-mo !

pp-mo:
> I think we're all good, to the limit of my understanding! Well done @pp-mo !

Thanks so much for all your diligence.
Unlike the "daughter" PRs, you can merge this one, if you wanted ?

PR #824, opened 2026-08-06
Final 0v23v0

pp-mo:
Fix the mainpage version number


pp-mo:
Changes adopted elsewhere

